PR #10, opened 2017-10-26
Fix icon loading

dan-lee:
We had the problem that the X icon in the app was not loaded anymore.
Getting rid of `this.mounted` and `state` and using `this.forceUpdate()` circumvented that problem.
I am not sure if this has some other implications, can you check it @mosch?

### DIFF
--- a/src/atoms/Icon.jsx
+++ b/src/atoms/Icon.jsx
@@ -175,12 +175,6 @@ export default class Icon extends React.Component {
     children: PropTypes.node,
   }
 
-  state = {
-    icon: typeof Icon[this.props.name] !== 'undefined',
-  }
-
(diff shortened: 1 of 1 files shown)


fivenp:
at least visually - it works now as expected

mosch:
Hey @dan-lee, you know what the exact problem was?
We cannot call forceUpdate in that asyc function without checking that the component is still mounted. Because may not be anymore.

dan-lee:
Ok then we should add that back. I am not entirely sure what the problem.
I am suspecting it has to do with this line:
https://github.com/allthings/elements/blob/master/src/atoms/Icon.jsx#L230
`} else if (!this.state.icon) {`

PR #11, opened 2017-10-26
adding responsive behaviour to text in hero image

fivenp:
we need this for our 1-column microapps which whould have weird text-size on tablet/desktop views :)


fivenp:
just realised the MQ is of course causing issues on the 2 column layout - will fix :)

fivenp:
* Responsiveness in Hero is fixed and tested
* Right Arrow Icon is added to ChevronList

mosch:
hey @fivenp, thanks for bringing that back in!

PR #20, opened 2017-11-06
extending the docs

fivenp:
quirky tryouts which actually dont look thaaaat bad so far...


mosch:
Great work, thanks @fivenp 👍🏻
I'll try to see why the code some times is not loading...

dan-lee:
The code is actually loading but you need to scroll to the right. It seems the dimensions are not calculated correctly on the first render.

fivenp:
which might got introduced by me while adding the bootstrap col-md wrappers around the editor

PR #28, opened 2017-11-16
remove hover bg color

hasznosl:
the hover bg color can conflict with the text color. While adding the cursor pointer by default is a good idea, if the user of the component wants hover style then they should pass that in as part of the styles.


mosch:
I'm not sure if I agree with this. Imho Button hover colors should stay the same across all applications. @fivenp any thoughts on this?

hasznosl:
Hover should look good with the child color, child text should remain readable. If we want a fixed hover color, we also need fixed child color. How can we enforce that? In my opinion we shouldn't.

hasznosl:
here @mosch : https://allthings.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=APP&view=detail&selectedIssue=APP-650

CancelBookingButton

this is the pr: https://github.com/allthings/app/pull/1111

dan-lee:
Closed in favor of https://github.com/allthings/elements/pull/34

PR #47, opened 2017-12-05
Add FormValidityProvider & tests 🚀.

yamafaktory:
In order to make testing fun & easier, I pushed Enzyme into the Jest pipeline.

Checklist:


.eslintrc
```diff
@@ -30,10 +32,13 @@
     "react/jsx-no-target-blank": 0,
     "react/no-array-index-key": 2,
     "react/no-multi-comp": [1, { "ignoreStateless": true }],
-    "react/self-closing-comp": [2, {
-      "component": true,
-      "html": true
-    }],
+    "react/self-closing-comp": [
```
yamafaktory: Some noise coming from Prettier...

src/molecules/TextInput.jsx
```diff
@@ -121,7 +121,16 @@ class TextInput extends React.Component {
...
+    const { STATES, validity } = this.context
+    const message = e.target.validationMessage
+    let hasState = false
+    for (const state of STATES) {
+      if (e.target.validity[state]) {
+        this.setState({ message: validity[state] || message })
+        break
+      }
+    }
+    hasState && this.setState({ message })
```
mosch: I think this can `hasState` will never be true - do I miss something?

src/molecules/TextInput.jsx
```diff
@@ -121,7 +121,16 @@ class TextInput extends React.Component {
 
   handleInvalid = e => {
     e.preventDefault()
-    this.setState({ message: e.target.validationMessage })
+    const { STATES, validity } = this.context
```
mosch: Does this work if no context is provided at all?

src/molecules/TextInput.jsx
yamafaktory: Yeah, I forgot to add it... but it was in my mind :D

src/molecules/TextInput.jsx
yamafaktory: Should, let me test it!

yamafaktory: Fixed in last commit, added a test for it.

src/behaviour/FormValidityProvider.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+/**
```
mosch: For proper doc generating (check out yarn build:docs) I guess the docs need to be in the line before the class declaration see 👇🏻

src/behaviour/FormValidityProvider.jsx
```diff
@@ -0,0 +1,54 @@
...
+  'tooShort',
+  'typeMismatch',
+  'valueMissing',
+]
+
+const validityPropTypes = STATES.reduce((acc, prop) => ({
+  ...acc,
+  [prop]: PropTypes.string,
+}), {})
+
```
mosch: 👈🏽 here

yamafaktory: Done.

PR #80, opened 2018-01-12
Add cdn intl provider

mosch:
The CDNIntlProvider gets project translation files from CDN and makes them available to use

Checklist:


package.json
```diff
@@ -14,6 +14,7 @@
     "neue": "1.0.0",
     "prop-types": "15.6.0",
     "react-broadcast": "0.6.1",
+    "react-intl": "^2.4.0",
```
yamafaktory: Shouldn't we use a pinned version instead?

PR #99, opened 2018-01-30
PhoneInput element and country codes file added

kialysa:
PhoneInput element added
  - adds space after country code (if it exists)
  - formats area code (if a specific format exists for that country)

Country code file added
  - object of country name, country code, area codes (if easily available), phone number template (if easily available)

Checklist:


mosch:
Nice addition to elements  🎊it looks good to me. You think you could provide some documentation for this?

src/molecules/PhoneInput.jsx
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * PhoneInputs are used to enter phone numbers. Entered information will automatically be formatted according to the country code, where applicable.
+ * Currently country code is formatted for all countries. Area codes are formatted for Switzerland, France, and the United States.
+ *
+ * ```example
+ * <PhoneInput name="phone" required />
+ * ```
+ * ```
+ **/
+
```
mosch: you should drop that blank line _

dan-lee:
Looking great! Good work so far @kialysa!
One thing I noted is when I type in a number, set the cursor to the beginning and delete a number the cursor jumps right back to the end:

kialysa:
Thanks for catching that, @dan-lee ! I've fixed it. Please find more bugs 😆

dan-lee:
So now it's kind of the other way around 😄If I set the cursor to the beginning and start to enter a number it'll jump to the end. I don't think that's intended?

kialysa:
Thanks, @dan-lee . Should be fixed now.🛠 Let me know if you find more!

PR #105, opened 2018-02-05
Allow to pass background color for Checkbox and TextInput

mosch:
**This changes the background color behaviour introduced in https://github.com/allthings/elements/pull/95**

In #95 the default behavior of Checkbox was changed. It takes the background color from the theme as background color of the ListItem the Checkbox is using. The background color of the theme is also used for the **background the UI elements are placed** on. 

So the change of #95 get's us this default behaviour: 

screen shot 2018-02-05 at 13 03 14

Only the Checkbox form element now has the default behaviour of having the same color as the background, and this cannot be changed. And changes the behavior of all existing checkboxes. 

I assume this is not what we actually want. My changes allow to pass a _backgroundColor_ as  a prop to the Checkbox and also the TextInput, and they get just passed to the ListItem.

Checklist:

### DIFF
--- a/src/molecules/Checkbox.test.js
+++ b/src/molecules/Checkbox.test.js
@@ -3,7 +3,7 @@ import Checkbox from './Checkbox'
 import { ThemeProvider } from '../'
 
 const THEME = {
-    background: 'darkMarco',
+  background: 'darkMarco',
 }
 const STRING_LABEL = 'I am a nice checkbox!'
 
@@ -29,7 +29,7 @@ describe('Test the checkbox component', () => {
   it('should work with children', () => {
     const wrapper = mount(
       <ThemeProvider theme={THEME}>
-        <Checkbox label={<h1>test</h1>} name="a" />
+        <Checkbox label={<h1>test</h1>} name="a" backgroundColor="darkMarco" />
       </ThemeProvider>
     )
     expect(wrapper).toMatchSnapshot()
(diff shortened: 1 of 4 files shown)


yamafaktory:
@fivenp mentioned this issue when we talked about my previous changes...

src/molecules/Checkbox.test.js
```diff
@@ -29,7 +29,7 @@ describe('Test the checkbox component', () => {
   it('should work with children', () => {
     const wrapper = mount(
       <ThemeProvider theme={THEME}>
-        <Checkbox label={<h1>test</h1>} name="a" />
+        <Checkbox label={<h1>test</h1>} name="a" backgroundColor="darkMarco" />
```
adieuadieu: Is `darkMarco` supposed to be an actual color hex code? It ends up in the CSS as ` background-color: darkMarco; ` which isn't a valid CSS named-color?

mosch: darkMarco is a special new color in the CSS4 spec, check the RFC. It's a dark, cruel and scary color. We like it.

yamafaktory: https://www.facebook.com/public/Marco-Dark

mosch:
yeah, @fivenp and @yamafaktory I think we need a documentation of colors and what they are "good" for. It's not too obvious just from code.

src/molecules/Checkbox.test.js
dan-lee: darkMarco is my favorite color. I have it as a wallpaper at home

yamafaktory: I've sent a request... https://drafts.csswg.org/css-color/#named-colors

adieuadieu: :trollface:

PR #107, opened 2018-02-06
Add stage to the CDNIntlProvider.

yamafaktory:
- Either `prodution` (default) or `staging`.

Checklist:

### DIFF
--- a/src/behaviour/CDNIntlProvider.jsx
+++ b/src/behaviour/CDNIntlProvider.jsx
@@ -9,7 +9,7 @@ class CDNIntlProvider extends React.Component {
     messages: PropTypes.object,
     onDone: PropTypes.func,
     project: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


src/behaviour/CDNIntlProvider.jsx
```diff
@@ -5,16 +5,18 @@ import { IntlProvider } from 'react-intl'
 class CDNIntlProvider extends React.Component {
   static propTypes = {
     children: PropTypes.node,
-    messages: PropTypes.object,
     locale: PropTypes.string.isRequired,
+    messages: PropTypes.object,
     onDone: PropTypes.func,
     project: PropTypes.string.isRequired,
+    stage: PropTypes.oneOf(['production', 'staging']),
```
adieuadieu: `prerelease` may in the future also be valid?

yamafaktory: ok

PR #124, opened 2018-02-19
Add fetchAlternative prop.

yamafaktory:
This is especially useful with SSR when an alternative fetch method can be used.

Checklist:

### DIFF
--- a/src/behaviour/CDNIntlProvider.test.jsx
+++ b/src/behaviour/CDNIntlProvider.test.jsx
@@ -100,7 +100,7 @@ describe('Check the CDNIntlProvider component', () => {
     expect(wrapper.props().stage).toBe(stage)
   })
   it('should use an alternative fetch method', async () => {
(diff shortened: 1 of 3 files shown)


src/behaviour/CDNIntlProvider.test.jsx
```diff
@@ -99,4 +102,34 @@ describe('Check the CDNIntlProvider component', () => {
...
+    )
+    const testRenderer = await new Promise((resolve, reject) => {
+      let myRenderer
+      function onDone() {
+        resolve(myRenderer)
+      }
+      const nbm = (
+        <ResourceProvider>
+          <CDNIntlProvider
+            fetchAlternative={fetchMethod}
```
dan-lee: @yamafaktory I guess you forgot to change the prop name?

yamafaktory: Fixed in #125

PR #131, opened 2018-02-21
adding buttons sizes and cursor:pointer to floatingbutton

fivenp:
Adding the possibility to pass different button sizes to this component.

The only culprint - the `<Text>` component which is usually used in combination with this button doesn't scale automatically the font-size (aka. you need to add `<Text size="l">` if you want a `<FloatingButton size="xl">`)

Checklist:


src/molecules/FloatingButton.jsx
```diff
@@ -15,18 +15,31 @@ const buttonStyle = css({
...
 class FloatingButton extends React.Component {
   static propTypes = {
     color: PropTypes.string.isRequired,
     disabled: PropTypes.bool,
     disabledColor: PropTypes.string.isRequired,
+    size: PropTypes.oneOf(['xs', 's', 'm', 'l', 'xl']),
+  }
+
+  static defaultProps = {
+    size: 'l',
```
fivenp: I'm not sure if we want to switch the default to large 

* Before this PR the FloatingButton is set to M (aka 50px) across all implementations (Booking, Business Directory, etc)
* We realized 50px is too small - so we'd like it to be 60px everywhere but aren't sure if its a good approach to simply default to L without knowing the real implications caused by this step...

fivenp: I did some research aaaand:

#### current usage of the `<FloatingButton>` seems to be like

```jsx
<FloatingButton>
  <Text strong size="s">text</Text>
<FloatingButton>
```
* FloatingButton is **fixed** to 50px height
* Text is 12px (set by size="s")

#### which would translate after this PR into
```jsx
<FloatingButton size="m">
  <Text strong size="s">text</Text>
<FloatingButton>
```
* FloatingButton is 50px heigh (set by size="m")
* Text is 12px (set by size="s")

#### perfect final usage would be
```jsx
<FloatingButton>
  <Text strong>text</Text>
<FloatingButton>
```
* FloatingButton **defaults** to 60px height
* Text **defaults** to 13px

#### proposal for a smooth transition:
* Implement this PR and default to size "l" for the FloatingButton
* Do a console.warn telling people to make sure to update their `<Text>` child within the `FloatingButton` to remove the `size` prop if it's set at all....

mosch: The FloatingButton hasn't too much use cases yet, so I think we can just change it without the console.log. Imho it's not a breaking change but a desired design change anyway.

fivenp: Hard to say - but I have some use use cases in mind where we might want to have at least xs, l, and xl FloatingButtons

fivenp:
@mosch - how shall I add the correct console.warn as proposed in https://github.com/allthings/elements/pull/131#discussion_r169915788 ?

I think it should be good to go after implementing this transition info

mosch:
Why not just increase the default from 50px to 60px for now, and if we need more sizes, we add them then?

mosch:
@fivenp ping, should we proceed or close this PR?

fivenp:
Proceeding of course :)

mosch:
Closed due inactivity

Issue #118, opened 2018-02-16
Horizontal transition should scroll up 


adashst:
When the horizontal transtition due to a page-change is started while scrolled down to the bottom, the view stays at the bottom (see attached mov).
Maybe we should scroll up in this case?
  
[Untitled.mov.zip](https://github.com/allthings/elements/files/1732180/Untitled.mov.zip)

mosch:
Hi @adashst, thanks for bringing this up. 
Yes, we could fix this by scrolling up automatically when a new screen is added. It may also work if you move scrolling to the containers instead of doing it in the HorizontalView.

What you have right now:

```

                               HorizontalView

+-------------------------------------------------------------------------------------+
|                                                                                    ||
|            A                           B                            C              ||
|   +---------------------+   +----------------------+    +-----------------------+  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   |                     |   |                      |    |                       |  ||
|   +---------------------+   +----------------------+    +-----------------------+  ||
|                                                                                    ||
|                                                                            +-----> ||
+-------------------------------------------------------------------------------------+
                                                                             |
                                                                             |
                                                                             |
                                          Scrolling in Container             +

```

**What could fix it:**

```
           +  Scrolling in Subviews
           |
           |                    HorizontalView
           |
 +-------------------------------------------------------------------------------------+
 |         |                                                                           |
 |         |  A                           B                            C               |
 |   +---------------------+   +----------------------+    +-----------------------+   |
 |   |     |              ||   |                     ||    |                      ||   |
 |   |     |              ||   |                     ||    |                      ||   |
 |   |     +--------------------------------------------------------------------> ||   |
 |   |     |              ||   |                     ||    |                      ||   |
 |   |     |              ||   |                     ||    |                      ||   |
 |   |     +-------------------------------------->  ||    |                      ||   |
[...]

PR #144, opened 2018-02-27
Update HorizontalView.jsx

adashst:
Simply scroll to 0,0 if scrollY prop is provided , and `window.pageYOffset` is > 0
```
<HorizontalView scrollY>
...
</HorizontalView>
```
Solves https://github.com/allthings/elements/issues/118

Checklist:

- [] Documentation added / updated

### DIFF
--- a/src/behaviour/HorizontalView.jsx
+++ b/src/behaviour/HorizontalView.jsx
@@ -49,6 +49,7 @@ import { css } from 'glamor'
 class HorizontalView extends React.Component {
   static propTypes = {
     children: PropTypes.array.isRequired,
+    scrollY: PropTypes.bool,
   }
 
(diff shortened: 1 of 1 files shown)


mosch:
Hey @adashst! What's your use-cases for this? When I get it right, you are scrolling up the main window when you create this component? I'm not sure if this is a constructive behaviour.

adashst:
@mosch Well, i added a small screencast here: https://github.com/allthings/elements/issues/118 to visualize my use-case.
But now i know what you meant with your suggestion. Of course its better to make the header sticky, and scroll just the content and not the complete view.
Will try that and see if it solves the scrolling issue on route-change.
So i'm going to close this.
Thanks!

PR #145, opened 2018-02-27
Update NotificationBubble.jsx

adashst:
Add custom delay if the user needs more time to read the NotificationBubble text.

Checklist:

- [X] Documentation added / updated


mosch:
Hi @adashst, thanks for the contribution! I like the idea of having more than just the default timeout for the NotificationBubble. I've got to additions to that: 

- Maybe the default is a bit too short (Material UI uses 2750ms, maybe we can just adapt this)
- To get a consistent behaviour I think it be good to provide predefined values ( default, long, extra-long?) instead of giving out the full control to user-land.

Any opinions on that?

src/molecules/NotificationBubble.jsx
```diff
@@ -72,7 +99,10 @@ class NotificationBubble extends React.Component {
   }
 
   componentDidMount() {
-    setTimeout(() => this.setState({ visible: false }), 2000)
+    setTimeout(
+      () => this.setState({ visible: false }),
+      this.props.delay || 2000
```
dan-lee: Rather use a defaultProp for delay makes it more obvious.

adashst:
@dan-lee Thanks for your hint. Of course defaultProps is the way to go.
@mosch I also felt that 2000ms is quite too short, just want to retain it. But why exactly 2750ms? 🤔

I like the idea of having some predefined values. On the other hand i think its very subjectively.
Maybe 2750ms is too short for one, while 3500ms seems to much...
So why not give one the freedom to define the delay that exactly fits their need, if the default of 2750 does not.

src/molecules/NotificationBubble.jsx
adashst: Thanks for your hint. Of course defaultProps is the way to go.

mosch:
I don't know why it's 2750ms. I found this in Material UI specs, so I guess Googles design team tested this with hundreds of users to find the perfect value. Or maybe it's just random 😅

I would predefine the value _because_ it's so subjectively and everyone would use other values. So we end up with a different behaviour for every Bubble. I assume in the end for the users it's best if all Bubbles behave kind of the same, and they will _"know"_ how long it will stay.

mosch:
@monsterlie @fivenp some ux opinions on that?

monsterlie:
I would also stay with on defined time. Just to have a consistency and as you said @mosch the user will get used to it and if the behavior is different for each bubble it might be annoying and/or confusing.

adashst:
To clarify. I didn't vote for different behaviour on each bubble. I agree thats its good to have a consistent behaviour. I just voted for having the freedom to define the time myself _if the default of 2750ms doesn't fit_
But i'm ok if we define a set of delays like: short, default, long, x-long.

mosch:
So let's start with _default_ and _extended_ then, so we have two values to choose from. Good?

adashst:
So i set the delay to 2750ms and removed the custom delay. 🙂

mosch:
🤛🏼 cool!

PR #152, opened 2018-03-06
Instructions for local development

dan-lee:
I added a section in the README one can follow to have an easier time developing with elements locally in their app.

Can you review if it works on your machine and check the wording? Thanks 🙏


dan-lee:
Also /cc @adashst, maybe you could make use of this?

fivenp:
Works like a charm! 💘

PR #154, opened 2018-03-07
Improved form elements

mosch:
We've added the label feature to TextInput, and it looks like that: 

jolo

Checklist:


mosch:
@HorstMumpitz might be cool for you too!

PR #155, opened 2018-03-07
feature: remove TextInput border-bottom with hideLine prop

kialysa:
Currently, TextInput and PhoneInput component automatically has a border-bottom due to ListItem.
The changes will allow the passing of hideLine prop to trigger ListItem hideLine.

Checklist:


src/molecules/TextInput.jsx
```diff
@@ -160,7 +160,11 @@ class TextInput extends React.Component {
     return (
       <Theme>
         {({ theme, colorize }) => (
-          <ListItem padded={false} backgroundColor={colorize(backgroundColor)}>
+          <ListItem
+            padded={false}
+            backgroundColor={colorize(backgroundColor)}
+            {...props}
```
dan-lee: Be careful with spreading all the props, the same props are also spread on the input field. Now the div has a `placeholder`. (see snapshot)

src/behaviour/__snapshots__/FormValidityProvider.test.jsx.snap
```diff
@@ -180,6 +180,9 @@ exports[`Check FormValidityProvider should click on the button and get a custom
   <div
     data-css-dau5wj=""
     data-css-h7582n=""
+    name="email"
+    placeholder="Your email"
```
dan-lee: ^

src/molecules/TextInput.jsx
kialysa: yikes. thanks for pointing that out!

mosch:
What's your actual need for hiding the line? I think putting the TextInput into a ListItem was not the best decision in the first place. So maybe it's the right moment to split things apart?

If not, I would go for "withLine" or "showLine" (default true) to get to consistent "always positive" prop-types.

mosch:
See: #156

mosch:
closing this in fav of https://github.com/allthings/elements/pull/156

PR #156, opened 2018-03-08
move input to atom, decouple from ListItem

mosch:
Checklist:

### DIFF
--- a/src/atoms/Input.jsx
+++ b/src/atoms/Input.jsx
@@ -179,7 +179,7 @@ class TextInput extends React.Component {
   handleMessageClick = () => this.setState({ message: null })
 
   render() {
-    const { required, backgroundColor, lines, label, ...props } = this.props
+    const { required, lines, label, ...props } = this.props
     const currentValue = this.props.value || this.state.value
     const labelVisible = currentValue.length > 0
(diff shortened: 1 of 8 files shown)


src/atoms/Text.jsx
```diff
@@ -14,7 +14,7 @@ if (typeof window !== `undefined`) {
   ])
 }
 
-const availableSizes = {
+export const availableSizes = {
```
dan-lee: Do we need this to be exported?

src/atoms/Input.jsx
```diff
@@ -0,0 +1,259 @@
...
+      message: null,
+      length: e.target.value.length,
+    })
+    this.props.onChange && this.props.onChange(e)
+  }
+
+  handleMessageClick = () => this.setState({ message: null })
+
+  render() {
+    const { required, backgroundColor, lines, label, ...props } = this.props
```
dan-lee: If we get rid of this, the tests should pass.

dan-lee:
The gist of this is moving out the input element into `Input` atom component and leaving the `TextInput` basically the same, right?

mosch:
exactly!

PR #162, opened 2018-03-08
Atom input phone number fix

mosch:
Checklist:


mosch:
@kialysa let's work on this 🙈

mosch:
@kialysa yay! Do you think we are good now?

PR #171, opened 2018-03-08
Add story for ColumnLayout

dan-lee:
I am not sure how to put this cleanly as an example so we can start having this as a story.


fivenp:
@dan-lee - I guess we should also update the sample 2-column-app in the [Example repo](https://github.com/allthings/elements-example/tree/master/two-column-responsive-layout)?

dan-lee:
Sorry didn't know about that. I think we can close this PR then.

Issue #176, opened 2018-03-13
Double rendering because of CDNIntlProvider mismatch


dan-lee:
When not passing the messages on client side but having them on server side a mismatch happens, and the whole app renders twice.
> Warning: Expected server HTML to contain a matching \<div\> in \<div\>

In my opinion this should not happen. 

Text mismatches still happen but these do not cause a re-render of the whole app, e.g.:
> Warning: Text content did not match. Server: "Hang on..." Client: "Hold on..."

We simply could remove `this.state.loaded` to change this behaviour:
https://github.com/allthings/elements/blob/2b86f249708ded2af4e84afba0f978360e359cf9/src/behaviour/CDNIntlProvider.jsx#L91-L97

mosch:
Hey @dan-lee, good catch! I think we should not render everything if it's not loaded yet. 
We can pass in messages to the component and it should be in the loaded state right away. So we I see two options there:
 1. Load messages using `loadLanguage` before you hydrate on client side and pass them directly to the IntlProvider
 2. Render messages into the DOM when you do SSR, so the IntlProvider can pick them up from there without any delay

Let's find the best option and add this to the docs!

dan-lee:
In my opinion we should render *something* to prevent the mismatch and eventually prevent a complete remount of the app. I'd prefer flashing text messages of everything disappearing and appearing again.

Of course this *shouldn't* happen when the messages are rehydrated from the server. But in some cases this may not happen (for some reason) and we would prevent a remount, virtually flashing the whole page.

I like your idea mentioned in the last paragraph, but I am not sure if this can work out. We can give it a try 👍

dan-lee:
Fixed with #178

PR #178, opened 2018-03-14
Prevent double render of CDNIntlProvider

dan-lee:
This introduces a new HTML element which renders differently on client and server side (with [hydrate](https://reactjs.org/docs/react-dom.html#hydrate) in mind) to rehydrate the messages.
This needs to happen in the constructor because the IntlProvider implementation does not behave well on prop changes (see https://github.com/yahoo/react-intl/issues/1097#issuecomment-363114792)

Discussion in #176 

Checklist:


mosch:
nice work! the container might not be there if we don't use ssr, so we also need to handle that.

mosch:
I added the tests, looks good :-)

PR #180, opened 2018-03-16
Input component improvements

dan-lee:
Several improvements:

* Just animate `padding-top` (`outline` was also animated).
* Render label only when provided. No text with an asterisk doesn't look _too_ good.
* If no `pattern` is provided don't bother to show the checkmark for validity.
* Remove default placeholder (remaining artefact?)
* Added basic prop tests
* ~~Make it possible to pass an error message, necessary for server side errors.~~ _(removed due increasing complexity and not fully spec'd)_

Checklist:


dan-lee:
Good now @mosch?

mosch:
@dan-lee good to merge?

PR #202, opened 2018-04-12
Input checkmark & autofill fixes

dan-lee:
**What has changed?**
* Show checkmark not only when `pattern` is passed but also when `minLength`, `maxLength` or `required` is passed.
* `&:-webkit-autofill ~ div` selector also selected the checkmark and applied `top: 8px` on it. This is solved by using class names on label/checkmark
* Use `opacity` on the checkmark to display/hide it. With this we can use the same technique to display the checkmark when it was autofilled ✌️
* Bonus: pass `required` to `textarea` as per spec it's a valid attribute.

Checklist:

### DIFF
--- a/src/atoms/Input.jsx
+++ b/src/atoms/Input.jsx
@@ -62,6 +62,15 @@ const styles = {
     borderRight: '10px solid transparent',
     borderTop: '10px solid #c1392b',
   }),
+  checkmark: show =>
+    css({
+      position: 'absolute',
+      top: 16,
+      right: 15,
+      pointerEvents: 'none',
+      transition: 'opacity .225s',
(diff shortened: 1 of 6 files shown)


src/atoms/Input.jsx
```diff
@@ -195,6 +207,12 @@ class Input extends React.Component {
     const labelVisible = currentValue.length > 0
     const showLabel = label && currentValue.length > 0
 
+    const isCheckmarkActive =
+      (pattern || props.minLength || props.maxLength || required) &&
```
dan-lee: Should we show the checkmark when `required` is satisfied or rather leave it out, what do you think?

PR #207, opened 2018-04-19
Remove fetch polyfill

dan-lee:
We decided that we, as a library, should not be responsible for providing a working `fetch` polyfill. It has to be provided from the apps consuming `@allthings/elements`.

Checklist:


dan-lee:
@mosch @yamafaktory This probably needs a v2, what do you think?

mosch:
@dan-lee agree. 
does anyone has breaking changes in mind we wanted to do also? we could use the chance to put them in one version.

PR #225, opened 2018-05-22
Radio button set

adashst:
RadioButtonSet can be used to render a set of RadioButtons to allow users to select exactly one item from a set of values.

Checklist:

- [X] Test added / Snapshots updated
- [X] Documentation added / updated


src/molecules/RadioButtonSet.jsx
```diff
@@ -0,0 +1,163 @@
...
+      defaultValue,
+      onChange,
+      labelSize,
+      ...props
+    } = this.props
+
+    return (
+      <Theme>
+        {({ theme, colorize }) => (
+          <ListItem
```
dan-lee: We made the mistake wrapping `CheckBox` and `Input` with a `ListItem`, which might not be expected. I would recommend removing the `ListItem` and leave it as the consumers concern how to wrap it.

src/molecules/RadioButtonSet.jsx
```diff
@@ -0,0 +1,163 @@
...
+
+const styles = {
+  radioWrapper: () =>
+    css({
+      padding: '15px 15px',
+    }),
+  radioElement: num =>
+    css({
+      marginRight: num > 2 ? 0 : 40,
+      width: `${num > 2 ? '100%' : 'auto'}`,
```
dan-lee: No need to wrap it in a string literal, you can just do: `width: num > 2 ? '100%' : 'auto'`.

mosch:
@adashst just added the prop-type for required. Docs failed because it was missing.

mosch:
👏🏼

PR #282, opened 2018-06-29
Typeahead

hasznosl:
Checklist:


src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
...
+import PropTypes from 'prop-types'
+import { View, ListSpinner, Text, ListItem, Icon } from '../'
+import { css } from 'glamor'
+import { ColorPalette } from '@allthings/colors'
+
+class Typeahead extends React.Component {
+  static propTypes = {
+    configuration: PropTypes.object,
+    width: PropTypes.string,
+    selectCallback: PropTypes.func,
```
mosch: this should be called "onSelect"

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { View, ListSpinner, Text, ListItem, Icon } from '../'
+import { css } from 'glamor'
+import { ColorPalette } from '@allthings/colors'
+
+class Typeahead extends React.Component {
+  static propTypes = {
+    configuration: PropTypes.object,
```
mosch: Try to use PropTypes.shape to describe complex configurations.

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
...
+    width: PropTypes.string,
```
mosch: instead of having a width prop, you could simply allow to pass css like the other components do with the rest props

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
...
+    selectCallback: PropTypes.func,
+    // deleteElement: PropTypes.func,
+    multiselect: PropTypes.bool,
+    nrResults: PropTypes.number,
```
mosch: There is a better name for this. Actually, I think this can be opinionated by the component, i don't really see a need to make it configurable.

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
...
+  componentDidMount() {
+    document.addEventListener('mousedown', this.handleClickOutside)
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('mousedown', this.handleClickOutside)
+  }
+
+  refetch = input => {
+    this.setState({ loading: true })
```
mosch: Instead of this, you need to have a property that the implementation can use to let the component know the results. S.th. like "onRequestResults()" and fetching and returing the list will be up to the implementation.

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,241 @@
...
+          <View
+            style={{
+              position: 'absolute',
+              zIndex: '2200',
+              cursor: 'pointer',
+              top: 18,
+              right: 10,
+            }}
+          >
+            <Icon
```
mosch: Not sure if we need this actually. We don't use a clear Icon anywhere else, is there a good reason to start this here?

stories/CollapsibleStory.jsx
```diff
@@ -0,0 +1,178 @@
+import React from 'react'
```
mosch: Did this file slip into the PR? This seems unrelated to me.

hasznosl: I did a small refactoring on the storybook index and this refactoring included another component's story

src/organisms/Typeahead.jsx
fivenp: I actually like the clear icon idea... 
In this case it makes even more sense but it would also be a good addition for TextInput in the future...

fivenp: Just take care with the zIndex... It might be too high (even staying on top of modals although the component is NOT inside the modal)

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,259 @@
...
+  handleClickOutside = e => {
+    if (this.listRef.current && !this.listRef.current.contains(e.target)) {
+      this.setState({ showResults: false, focusIndex: -1 })
+    }
+  }
+
+  clearInput = () => this.setState({ input: '' })
+
+  onKeyDown = e => {
+    e.stopPropagation()
```
dan-lee: Consider [`Event.preventDefault`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault) on arrow keys instead of stopping the propagation.

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,267 @@
...
+            />
+          </View>
+        )}
+        <input
+          onChange={this.handleInputChange}
+          onFocus={this.onInputFocus}
+          value={input}
+          ref={this.inputRef}
+          tabIndex={tabIndex}
+          {...css({ height: 43, textIndent: 22, border: 'none' })}
```
fivenp: See above... If we don't use our own `<TextInput>` we should apply the same styles (fonts, textIndent:15, etc etc) to this `<input>`

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,267 @@
...
+                      this.handleSelectItem({
+                        value: option[configuration.value],
+                        label: option[configuration.label],
+                      })
+                    }
+                  }}
+                  {...css({
+                    width,
+                    backgroundColor: isSelected
+                      ? ColorPalette.lightGrey
```
fivenp: ` alpha(ColorPalette.background.light, 0.25)`

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,267 @@
...
+                      : ColorPalette.white,
+                    ':focus': {
+                      backgroundColor: isSelected
+                        ? ColorPalette.lightGreyIntense
```
fivenp: `alpha(ColorPalette.background.light, 0.40)`

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,267 @@
...
+                        value: option[configuration.value],
+                        label: option[configuration.label],
+                      })
+                    }
+                  }}
+                  {...css({
+                    width,
+                    backgroundColor: isSelected
+                      ? ColorPalette.lightGrey
+                      : ColorPalette.white,
```
fivenp:
```
':hover': {
  backgroundColor: alpha(ColorPalette.background.light, 0.25)
},
```

PR #284, opened 2018-07-04
HorizontalView: children.map to React.Children.map

kialysa:
Before: required to pass in an array of components. if last component is false, renders nothing

After: array is not required. if any component is false, it is filtered out

Checklist:

### DIFF
--- a/src/behaviour/HorizontalView.jsx
+++ b/src/behaviour/HorizontalView.jsx
@@ -110,7 +110,7 @@ class HorizontalView extends React.Component {
           onTransitionEnd={this.handleTransitionEnd}
           {...props}
         >
-          {children.map((child, i) => (
+          {React.Children.map(children, (child, i) => (
             <View
               // eslint-disable-next-line
               key={i}
--- a/src/behaviour/HorizontalView.jsx
+++ b/src/behaviour/HorizontalView.jsx
@@ -65,22 +65,24 @@ class HorizontalView extends React.Component {
     }
   }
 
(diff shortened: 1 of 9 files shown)


src/behaviour/HorizontalView.jsx
```diff
@@ -65,22 +65,24 @@ class HorizontalView extends React.Component {
     }
   }
 
-  componentWillReceiveProps(nextProps) {
-    const oldChildren = this.props.children
-    const nextChildren = nextProps.children
+  static getDerivedStateFromProps(
+    { children: nextProps },
+    { children: previousState }
+  ) {
```
kialysa: is nextProps.children or previousState.children ever undefined?
is this better?
```
  { children: nextProps = [] },
 { children: previousState = [] }
```

mosch:
Sweet! I think it would make sense to add a unit test for this. That would help preventing that this pops up again.

src/behaviour/HorizontalView.jsx
mosch: Not sure actually. I think the only case where it could be undefined is if you do <HorizontalView /> like this.

kialysa:
@mosch  you got it 👍

src/behaviour/HorizontalView.jsx
kialysa: Ok, then I'll leave it.

kialysa:
@mosch i've written the tests now! let me know if there's sth im missing

PR #286, opened 2018-07-05
Fix docs

mosch:
I fixed the broken docs introduced with the Collapsible.


src/molecules/Form/Form.jsx
```diff
@@ -116,6 +116,13 @@ class Form extends React.Component {
     onSubmit: _ => _,
   }
 
+  constructor(props) {
+    super(props)
+    this.form = React.createRef()
```
dan-lee: Can you put it as class variable instead?

src/molecules/Form/Form.jsx
```diff
@@ -116,6 +116,13 @@ class Form extends React.Component {
...
+  }
+
+  submit = () => this.form.submit()
```
dan-lee: This should be `this.form.current.submit()`, right?

PR #288, opened 2018-07-06
Allow fine-grain control of the height by the children.

yamafaktory:
The aim of this PR is to prepare the future integration of the Typeahead component inside a Collapsible one.

The current implementation relies on the height, i.e. is directly modifying the height of the children. This is an incorrect behavior when the aforementioned children are prone to have a fluctuating height.

The solution is to use a combination of height / min-height / max-height in order to keep the actual behavior while allowing more flexibility.

Checklist:


src/molecules/Collapsible.jsx
```diff
@@ -85,19 +85,23 @@ import { ColorPalette, alpha } from '@allthings/colors'
...
-    initiallyCollapsed: PropTypes.bool,
-    hasBottomBorder: PropTypes.bool,
-    tabIndex: PropTypes.number,
-    toggleCallback: PropTypes.func,
+    children: node,
+    hasBottomBorder: bool,
+    initiallyCollapsed: bool,
+    tabIndex: number,
+    title: string,
+    toggleCallback: func,
```
hasznosl: I don't prefer destructuring here, because if I want to add a new prop, I have to go above where it's imported, and add another type there too. It's faster to me to just copy the above line and change the type.

yamafaktory: Adjusted.

src/molecules/Collapsible.jsx
```diff
@@ -85,33 +86,73 @@ import { ColorPalette, alpha } from '@allthings/colors'
...
+      // height in realtime.
+      // See: https://caniuse.com/#feat=mutationobserver
+      // TODO: Would be neat to implement a nice `withObserver` HOC component if
+      // we plan to reuse such a nicety!
+      const MutationObserver =
+        window.MutationObserver ||
+        window.WebKitMutationObserver ||
+        window.MozMutationObserver
+      if (MutationObserver) {
+        this.observer = new MutationObserver(mutations => {
```
dan-lee: I am not sure if we should go on with this. We could do it similar to Material UI expansion panel: https://material-ui.com/demos/expansion-panels/

It changes right to `height: auto` after the transition is done. So there is no need to set the height again and cause an expensive layout reflow.

yamafaktory:
@dan-lee found an elegant solution for my current issue with #290.

However, we still to figure out if this still needed / valid for some corner-cases (child content / height changing afterwards).

PR #291, opened 2018-07-10
Collapsible: allow dynamic height

dan-lee:
Checklist:


src/molecules/Collapsible.jsx
```diff
@@ -212,10 +224,11 @@ class Collapsible extends React.Component {
         {/* Child */}
         <View
           onRef={this.childRef}
+          onTransitionEnd={this.handleTransitionEnd}
```
yamafaktory: Neat!

yamafaktory:
Well, this a way simpler and robust implementation performance-wise than #288! Closes #288.

yamafaktory:
@dan-lee @mosch I would like to reopen that PR or at least we should try to find an alternative solution for the dynamic height of the content. This is still an issue with the current implementation.

dan-lee:
@yamafaktory Sure, what's the issue? How to reproduce? I'd like to find a pure CSS fix if possible.

yamafaktory:
@dan-lee you can have a look at https://github.com/allthings/cockpit/pull/963. If you reach the detail page of any user (user management) and then, for a given channel, add more roles till those are overlapping the content of the collapsible. The collapsible component is not dynamically updating his height to its content.

dan-lee:
I don't think this PR is released yet. I can see, that the height is not changing to `auto` after collapsing/expanding.

PR #294, opened 2018-07-12
Remove CDNIntlProvider export from index

dan-lee:
Currently it's not possible to use the app without `react-intl` installed as the `CDNIntlProvider` is _always_ exported and as a result `react-intl` is _always_ imported.

On the other hand I would like to discuss if this `CDNIntlProvider` even belongs into `@allthings/elements` as I don't see any real value for 3rd parties here, what do you think?


yamafaktory:
That makes totally sense! On a similar level, we don't want React to be bundled so it's a peer dependency. Same already apply to react-intl.

mosch:
Good changeset! about 3rd parties: they are already making use of the "allthings translation service" and therefore I they should also be able to use this component. I would keep it.

PR #301, opened 2018-07-14
add story and do things performant

mosch:
Do it other ways: It's more reliable and performant.

Checklist:

### DIFF
--- a/src/layouts/SimpleLayout.jsx
+++ b/src/layouts/SimpleLayout.jsx
@@ -16,14 +16,13 @@ const simple = backgroundColor =>
     height: '100%',
   })
 
-const paddedCss = (paddedVertical, paddedHorizontal, additionalPaddingTop) =>
+const paddedCss = (paddedVertical, paddedHorizontal, pullDown) =>
   css({
+    transform: `translateY(${pullDown}px)`,
     [createMQ('tablet', 'desktop')]: {
       paddingLeft: paddedHorizontal && '25px',
       paddingRight: paddedHorizontal && '25px',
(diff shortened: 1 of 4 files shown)


dan-lee:
Add more ✨to the story please ✌️

mosch:
@kialysa @dan-lee I added a [simple story](https://deploy-preview-301--allthings-elements.netlify.com/) to showcase the "pull-to-refresh". It's simple but I think effective - what do you think?

src/layouts/SimpleLayout.jsx
```diff
@@ -16,14 +16,13 @@ const simple = backgroundColor =>
     height: '100%',
   })
 
-const paddedCss = (paddedVertical, paddedHorizontal, additionalPaddingTop) =>
+const paddedCss = (paddedVertical, paddedHorizontal, pullDown) =>
   css({
+    transform: `translateY(${pullDown}px)`,
```
dan-lee: I know it was like this before, but to make this even more performant I think it would make sense to use an inline style here. Because every percentage change will create a new CSS class and needs to go through the whole glamour process.

dan-lee:
One more thing I noticed when using it was when I was scrolling up quickly it already reloaded. I'll need to test this on a real device to see how sensitive it actually is.

kialysa:
@mosch this might be an obvious question: how do i "pull down" on the desktop? storybook isnt so great on mobile

mosch:
good point. you need to use your devtools to simulate mobile. not the best experience.

mosch:
@dan-lee ping

PR #312, opened 2018-07-25
Typeahead downshift

dan-lee:
Typeahead component using `downshift` and based on #282 

[See it in action](https://deploy-preview-312--allthings-elements.netlify.com/?selectedKind=Forms&selectedStory=Typeahead&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)


yamafaktory:
@dan-lee + @mosch I've implemented the changes we've talked about last week. I've extended the story for this new component accordingly. I've created a complete set of unit tests too and the docs have been generated.

@dan-lee I know we discussed that `clearOnSelect` prop but I already see some usage where we don't want a controlled or uncontrolled component anyway. So for that specific usage , we can now use that prop. Moreover, in order to avoid any possible confusion for the user, a warning is triggered if a `defaultValue` or `value` is used simultaneously (there's also a unit test).

src/organisms/Typeahead.jsx
```diff
@@ -0,0 +1,398 @@
...
+    onClose: NOOP,
+    onOpen: NOOP,
+  }
+
+  constructor(props) {
+    super(props)
+    if (
+      props.hasOwnProperty('clearOnSelect') &&
+      (props.hasOwnProperty('defaultValue') || props.hasOwnProperty('value'))
+    ) {
```
dan-lee: `&& IS_DEV` maybe?

PR #350, opened 2018-08-15
Add new behavior when using the clear icon ❎.

yamafaktory:
Checklist:

Live test: https://deploy-preview-350--allthings-elements.netlify.com/?selectedKind=Forms&selectedStory=Typeahead&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

### DIFF
--- a/src/organisms/Typeahead.jsx
+++ b/src/organisms/Typeahead.jsx
@@ -101,6 +101,8 @@ export default class Typeahead extends React.PureComponent {
   }
 
   clearSelection = downshiftClearSelection => () => {
+    // Focus back on the input.
+    this.inputRef && this.inputRef.focus()
     // Trigger the Downshift method.
     downshiftClearSelection()
     // Trigger the prop one.
@@ -126,6 +128,8 @@ export default class Typeahead extends React.PureComponent {
   }
 
   stateReducer = (state, changes) => {
+    const { clearOnSelect } = this.props
+
     switch (changes.type) {
       // Special case when the clearOnSelect property is used and we want to
       // clear the input.
@@ -133,7 +137,7 @@ export default class Typeahead extends React.PureComponent {
       case Downshift.stateChangeTypes.keyDownEnter:
         return {
           ...changes,
-          ...(this.props.clearOnSelect && { inputValue: '' }),
+          ...(clearOnSelect && { inputValue: '' }),
         }
 
       case Downshift.stateChangeTypes.changeInput:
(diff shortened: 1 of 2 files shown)


src/organisms/Typeahead.jsx
```diff
@@ -361,9 +382,8 @@ export default class Typeahead extends React.PureComponent {
                 {showOpen && (
                   <List
                     onRef={this.setListRef}
-                    {...getMenuProps({
-                      refKey: 'onRef',
-                    })}
+                    // Bypass the refKey check which is messy.
+                    {...getMenuProps({}, { suppressRefError: true })}
```
yamafaktory: This was the reason why the unit tests failed when bumping up the Downshift version. However, using `refKey` is kind of buggy and doesn't work as expected. Using the `suppressRefError` fixes the warning issue and solves the `onRef` usage (it was actually not working anymore...).

src/organisms/Typeahead.test.jsx
```diff
@@ -91,6 +91,8 @@ describe('Test the typeahead component', () => {
     expect(wrapper.find(CLEAR_SELECTION)).toHaveLength(1)
     wrapper.find(CLEAR_SELECTION).simulate('click')
     expect(wrapper.find(INPUT(0)).prop('value')).toBe('')
+    // The menu should be opened back for selection.
```
yamafaktory: A small test to be sure that it behaves (and will behave) as expected 🤞.

src/organisms/Typeahead.jsx
```diff
@@ -101,6 +101,8 @@ export default class Typeahead extends React.PureComponent {
   }
 
   clearSelection = downshiftClearSelection => () => {
+    // Focus back on the input.
```
yamafaktory: We need that focus back otherwise after clicking on the clear icon we are loosing the context and the user can use the keyboard interaction anymore 🐼.

src/organisms/Typeahead.jsx
```diff
@@ -187,6 +205,8 @@ export default class Typeahead extends React.PureComponent {
       showScrollArrow: this.listRef.scrollHeight > this.props.menuHeight,
     })
 
+  setInputRef = el => (this.inputRef = el)
```
dan-lee: As the min peer dep version of React is 16.3 can't we use `React.createRef`?

src/organisms/Typeahead.jsx
dan-lee: Oh what, this is strange :\. But if it works, we might check it again.

src/organisms/Typeahead.jsx
```diff
@@ -143,6 +147,20 @@ export default class Typeahead extends React.PureComponent {
           selectedItem: changes.inputValue === '' ? null : state.selectedItem,
         }
 
+      case Downshift.stateChangeTypes.unknow:
```
dan-lee: Typo, should be `unknown` I figure

src/organisms/Typeahead.jsx
yamafaktory: Yeah it works. If you don't use it (`suppressRefError`) and try with `refKey` the `onRef` doesn't work at all... and the unit tests completely fails because of the errors generated by Downshift.

src/organisms/Typeahead.jsx
yamafaktory: Good catch!

src/organisms/Typeahead.jsx
yamafaktory: This is a blocker I guess https://github.com/allthings/elements/blob/master/src/atoms/Input.jsx#L148 ?

yamafaktory:
@dan-lee please review it again 🚀, thanks a ton 🍻!

PR #354, opened 2018-08-16
Don't show the clear icon in cleared controlled component.

yamafaktory:
Checklist:

### DIFF
--- a/src/organisms/Typeahead.jsx
+++ b/src/organisms/Typeahead.jsx
@@ -94,8 +94,16 @@ export default class Typeahead extends React.PureComponent {
   componentDidUpdate({ value: prevValue = '' }) {
     // Force to coerce to an empty string when used as a controlled component.
     const { value = '' } = this.props
-    if (prevValue !== value && prevValue === '' && value !== '') {
-      this.setState({ forceShowClearIcon: true })
+
+    if (prevValue !== value) {
+      // Show the clear icon if controlled when we have a value.
+      if (prevValue === '' && value !== '') {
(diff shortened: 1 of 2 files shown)


src/organisms/Typeahead.jsx
```diff
@@ -94,8 +94,16 @@ export default class Typeahead extends React.PureComponent {
   componentDidUpdate({ value: prevValue = '' }) {
     // Force to coerce to an empty string when used as a controlled component.
     const { value = '' } = this.props
-    if (prevValue !== value && prevValue === '' && value !== '') {
-      this.setState({ forceShowClearIcon: true })
+
+    if (prevValue !== value) {
+      // Show the clear icon if controlled when we have a value.
+      if (prevValue === '' && value !== '') {
```
kialysa: just out of curiosity, in this `if` block,`preValue` and `value` could never be the same value, so it implies if `prevValue === ''` then `value` will never be `''`? So this could actually be written as 
```
if (prevValue === '') {...}
if (value === '') {...}
```
(am i making sense..? 🤯) Maybe it's more readable the way you wrote it though..🤔I don't think it's such a big deal so up to you! approved!

kialysa:
🔪🐛

src/organisms/Typeahead.jsx
yamafaktory: Good catch!

yamafaktory: Your hint headed me in the right direction, i.e. I discovered a better way to do the same with less pain! Thanks 👍.

PR #402, opened 2018-10-04
Spinner buttons

mosch:
This PR allows the floating button to indicate that something is in progress

screenshot 2018-10-04 at 09 47 34

Checklist:


src/molecules/SquareIconButton.jsx
```diff
@@ -63,8 +62,4 @@ class SquareIconButton extends React.Component {
   }
 }
 
-const mapThemeToProps = theme => ({
-  textColor: theme.text,
-})
-
-export default withTheme(mapThemeToProps)(SquareIconButton)
+export default SquareIconButton
```
fivenp: are you sure this doesn't do any harm?!
stuff like - "oh.. now the icon - color is dark on our AppMenu"

mosch: Imho it wasn't used at all – so I think this doesn't change anything?

fivenp: lets remove it and simply give it a try before upgrading :)

dan-lee:
Should we disable buttons which are marked in progress implicitly? I can't imagine a a use case where it should be in progress and still be clickable.

dan-lee:
Also Storybook → Button → in progress doesn't really show anything.

src/molecules/FloatingButton.jsx
```diff
@@ -36,13 +60,20 @@ class FloatingButton extends React.Component {
...
             height: 50,
             overflow: 'hidden',
             width: '100%',
           })}
         >
-          <button {...buttonStyle} {...props} />
+          <button
+            {...buttonStyle}
+            {...props}
+            onClick={inProgress || onClick}
```
dan-lee: ?

src/molecules/FloatingButton.jsx
```diff
@@ -13,17 +14,40 @@ const buttonStyle = css({
...
+    const {
+      color,
+      onClick,
+      disabled,
+      disabledColor,
+      inProgress,
+      children,
+      ...props
+    } = this.props
+    const isDisabled = disabled || inProgress
```
dan-lee: Please add a comment for why we need this.

PR #451, opened 2018-11-21
Confirm component

kialysa:
`Confirm` component to use in place of window.confirm()

Checklist:


src/molecules/ConfirmDialog.jsx
```diff
@@ -30,33 +30,33 @@ const styles = {
...
-      accept: PropTypes.string,
-      cancel: PropTypes.string,
-      message: PropTypes.node.isRequired,
-    }),
   }
 
   state = {
-    cancelMessage: this.props.customTexts.cancel || 'Cancel',
-    acceptMessage: this.props.customTexts.accept || 'OK',
+    cancelMessage: this.props.cancel || 'Cancel',
```
mosch: Two things about this: 
1. It doesn't need to go state since you can use defaultProps for that.
2. We don't provide Text defaults anywhere, so I'd just go for no default at all and make the Props required

src/molecules/ConfirmDialog.jsx
```diff
@@ -30,33 +30,33 @@ const styles = {
...
     flexDirection: 'column',
     alignItems: 'center',
-    backgroundColor: 'rgba(0, 0, 0, 0.5)',
+    backgroundColor: 'rgba(0, 0, 0, 0.25)',
   }),
 }
 
 class ConfirmDialog extends React.Component {
   static propTypes = {
+    accept: PropTypes.string,
```
mosch: maybe acceptButtonText is more accessible?

PR #507, opened 2018-12-10
introduced prop for typeahead to extend upwards

hasznosl:
Checklist:


stories/TypeaheadStory.jsx
```diff
@@ -134,6 +134,7 @@ class TypeaheadStory extends React.Component {
               onInputValueChange={this.onInputChange}
               items={movies}
               isLoading={loading}
+              extendUpwards={true}
```
yamafaktory: We should have another scenario in the story instead of changing that one.

yamafaktory:
@hasznosl Can you update the snaps please 🖼 📸 ?

stories/TypeaheadStory.jsx
hasznosl: thats at the bottom anyway, but yea I can do that

src/organisms/Typeahead.jsx
```diff
@@ -67,6 +82,7 @@ export default class Typeahead extends React.PureComponent {
     onClearSelection: NOOP,
     onClose: NOOP,
     onOpen: NOOP,
+    placement: 'bottom',
```
dan-lee:
```suggestion
    placement: Placement.bottom,
```

src/organisms/Typeahead.jsx
```diff
@@ -31,6 +44,8 @@ export default class Typeahead extends React.PureComponent {
     clearOnSelect: PropTypes.bool,
     /** The default value of the component, without making it controlled. */
     defaultValue: PropTypes.string,
+    /** If "top", then the list should be reversed and extended upwards, if "bottom" (default) then downwards */
+    placement: PropTypes.string,
```
dan-lee:
```suggestion
    placement: PropTypes.oneOf(Object.values(Placement)),
```

PR #584, opened 2019-01-28
Fix/button background color

fivenp:
I just realised `kewler` will break all our hero images / buttons where we pass `rgba()` or apparently even `rgb()` values with the last button update (PR #583)

Which makes partially sense... Since you cannot lighten/darken a transparent color (u basically can... but kewler can't :))

Soooo this is a hot fix to NOT break current implementations - it checks whether we pass a `rgb(a)` value (therefor no hover effect is added)

Checklist:

### DIFF
--- a/src/molecules/Button.jsx
+++ b/src/molecules/Button.jsx
@@ -34,9 +34,9 @@ function styles(
     ':hover': {
       background: disabled
         ? disabledBackgroundColor
-        : backgroundColor.includes('rgb')
-        ? backgroundColor
-        : col(backgroundColor, lightness(-10)),
+        : backgroundColor.includes('#')
+        ? col(backgroundColor, lightness(-10))
+        : backgroundColor,
     },
   })
 }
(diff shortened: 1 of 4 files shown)


dan-lee:
I think we should rather check for hex or non-hex, because there is not only `rgb/rgba` but also `hsl/hsla` and the whole HTML color palette.

fivenp:
> I think we should rather check for hex or non-hex, because there is not only `rgb/rgba` but also `hsl/hsla` and the whole HTML color palette.

You're totally right... Great point! My dream scenario is to also add `hsl()` and `rgb()` to `hex` conversions (no `hsla` and `rgba`) one day

dan-lee:
Cool! Btw: [No `includes` in IE](https://caniuse.com/#feat=es6-string-includes) 😕

fivenp:
> Cool! Btw: [No `includes` in IE](https://caniuse.com/#feat=es6-string-includes) 😕

Life is beautiful..... :)

PR #604, opened 2019-02-05
Replace `@babel/node` with `esm`

dan-lee:
I replaced [`@babel/node`](https://babeljs.io/docs/en/babel-node) with [`esm`](https://github.com/standard-things/esm). This is using default node with the `esm` runtime and provides much better debugging possibilities. In theory at some point we should be able to just remove `-r esm` 🤞

(also did some minor cleanup in the scripts)


dan-lee:
Unfortunately some scripts are not running smoothly since this change, so it's not done yet.

dan-lee:
`jest` now also runs on the `script` folder which had `example.test.js` inside, which actually never run.

PR #620, opened 2019-02-07
Customise Colors for Calendar

### LINKED ISSUE #602: Color for Calendar
The `Calendar` Component has it's active day color currently hardcoded - we should switch over to make it "dynamic"

fivenp:
This was intended to fix #602 one day - sadly it doesn't work that "easy"....

So for now - I just changed the disabled styling (otherwise it's hard to distinguish between disabled in the past and dates overlapping to the next month for example) and added ColorPalette colors which brigthen/darken accordingly :/

**Before**:
image

**After**:
image

Checklist:

### DIFF
--- a/src/molecules/Calendar.jsx
+++ b/src/molecules/Calendar.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import UnstyledCalendar from 'react-calendar/dist/entry.nostyle'
 import { ColorPalette } from '@allthings/colors'
 import insertCSS from '../utils/insertCSS'
+import { color, lightness } from 'kewler'
 
 insertCSS(`
 .react-calendar {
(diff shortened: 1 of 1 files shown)


mosch:
🤷🏼‍♂️

fivenp:
yeah.... I need to find some time to continue working on it :/

dan-lee:
Cool! I like the gradient pattern 🌈. But it looks a little bit weird when it's chopped off because the tiles end, doesn't it? I don't know if there's a good solution for that.

fivenp:
> Cool! I like the gradient pattern 🌈. But it looks a little bit weird when it's chopped off because the tiles end, doesn't it? I don't know if there's a good solution for that.

Fully agree... maybe we add a margin between the cells in future? :)

PR #621, opened 2019-02-07
Refactor build process

dan-lee:
This is to make the build/release process easier:

- Introduce `rollup` for bundling:
  - Thanks to `preserveModules: true` we can keep the folder structure. 
  - Generate a CJS and ESM bundle (before it was only CJS, even if the `package.json` stated otherwise)
- Use `yarn version` for creating releases and get rid of `yarn create:release`.

We should probably check with a pre-release version if it works correctly.


dan-lee:
Unfortunately this doesn't work in our current app setup 😔

mosch:
What's the matter? 🤷🏼‍♂️

dan-lee:
@mosch I can't remember exactly, but there were some problems. I think this would require a pair programming session, if you'd be up for it.

mosch:
Closed due to inactivity.

PR #626, opened 2019-02-11
Add TimeInput component

dan-lee:
[**Select a day first and then see it in action!**](https://deploy-preview-626--allthings-elements.netlify.com/?selectedKind=Containers&selectedStory=Filters&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

Checklist:


package.json
```diff
@@ -64,6 +64,7 @@
     "babel-plugin-add-react-displayname": "0.0.5",
     "babel-plugin-react-docgen": "2.0.2",
     "cross-env": "5.2.0",
+    "cross-fetch": "3.0.0",
```
dan-lee: For Storybook + IE 11 😇

src/molecules/TimeInput.jsx
```diff
@@ -0,0 +1,291 @@
...
+import { css } from 'glamor'
+
+import Input from '../atoms/Input'
+import View from '../atoms/View'
+import Absolute from '../atoms/Absolute'
+import Relative from '../atoms/Relative'
+import Text from '../atoms/Text'
+import Icon from '../atoms/Icon'
+
+const selectStyle = isActive =>
```
fivenp: We're most likely missing the font-family + size in here (or in a mother component?) - at least safari renders it "defaultish"

image

fivenp:
Wonderful stuff....

I guess it would be a feature request to use the 
`<input type="time" step="1500" /> <!-- 15 minutes -->` 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time on mobile / android only? :)

image

src/molecules/TimeInput.jsx
dan-lee: Thanks good catch. As we don't have global styles this probably needs `createTextStyles`

dan-lee:
@fivenp I started trying to use `<input type="time" />` on mobile iOS but this introduces more problems which would take some time to solve. So I would suggest moving on with it and try to improve after we've merged it.

fivenp:
> ...introduces more problems

absolutely true... stepping + max + min is not even supported.... :(

PR #627, opened 2019-02-11
DateInput: Call onChange when value is cleared

hasznosl:
Checklist:

For DateInput we the onChange will be called on clearing the field

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "test:example": "jest ./scripts/example.test.js",
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
-    "test": "jest src/ scripts/",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -38,7 +38,7 @@
     "test:example": "jest ./scripts/example.test.js",
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
-    "test": "jest src/ scripts/",
+    "test": "NODE_ENV=test jest src/ scripts/",
```
mosch: What do we need that for?

dan-lee: Our interal `devenv/init` sets `NODE_ENV=development` no matter what. Changing that could have a lot of side effects, so force it to `NODE_ENV=test` which it would do by default anyway.

PR #633, opened 2019-02-12
<form should default to POST method.

adashst:
This PR implements https://allthings.atlassian.net/browse/APP-2798.
Form method defaults to POST.

Checklist: 

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
     "mitt": "1.1.3",
     "neue": "1.0.0",
     "prop-types": "15.7.1",
-    "react-broadcast": "0.7.1",
     "react-calendar": "2.18.1",
     "react-media": "1.9.2",
     "react-motion": "0.5.2"
--- a/package.json
+++ b/package.json
@@ -37,8 +37,8 @@
     "test:example": "jest ./scripts/example.test.js",
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
-    "test": "NODE_ENV=test jest src/ scripts/",
-    "test:single": "NODE_ENV=test jest",
+    "test": "yarn jest src/ scripts/",
+    "jest": "NODE_ENV=test jest",
     "prettier": "prettier --no-semi --single-quote --trailing-comma es5 --write",
     "lint": "eslint --cache --ext .js,.jsx src",
     "storybook": "start-storybook -p 6006"
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
     "test": "NODE_ENV=test jest src/ scripts/",
+    "test:single": "NODE_ENV=test jest",
     "prettier": "prettier --no-semi --single-quote --trailing-comma es5 --write",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -38,6 +38,7 @@
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
     "test": "NODE_ENV=test jest src/ scripts/",
+    "test:single": "NODE_ENV=test jest",
```
dan-lee: What is this used for?

src/molecules/Form/Form.jsx
```diff
@@ -109,11 +111,13 @@ Also see the <a href="/molecules/TextInput/">TextInput</a> for allowed props.
 class Form extends React.Component {
   static propTypes = {
     children: PropTypes.node,
+    method: PropTypes.string,
     onSubmit: PropTypes.func.isRequired,
   }
 
   static defaultProps = {
     onSubmit: _ => _,
+    method: 'POST',
```
dan-lee: Couldn't this have any side effects on forms behaviour when submitting? I mean that it actually sends you to another page without `preventDefault` on form submission.

Check this CodeSandbox: https://codesandbox.io/s/92z69162qy

dan-lee: Oh never mind, our `handleSubmit` actually calls `preventDefault`. 👍

dan-lee:
As we call `preventDefault` in our submit handler it should never actually send the form data as GET parameters. Do we really need this? 🤔

package.json
adashst: I added this to test a single file.
If i do`yarn test Form.test` all tests are running.
With `yarn test:single Form.test`i can test just the `Form.test.js`

adashst:
@dan-lee https://allthings.atlassian.net/browse/APP-2546

dan-lee:
I saw that but couldn't figure out why it's a problem even with `preventDefault`.
But the discussion earlier cleared it up for me:
When the page is correctly loaded via SSR but the client code throws an error, the `preventDefault` is not triggered and default method `GET` for the form submission is used 💡

package.json
dan-lee:
```suggestion
    "jest": "NODE_ENV=test jest",
```

What about that? Executes the following for me:

>$ yarn jest src/molecules/FloatingButton.test.js 
>$ NODE_ENV=test jest src/molecules/FloatingButton.test.js

dan-lee: Or I think better would be:

```
"test": "NODE_ENV=test jest",
"test:all": "yarn test src/ scripts/"
```

Then you could still use `yarn test src/MyFile.jsx`

adashst: Think i like it.
So `yarn test` still behaves the same and `yarn jest Form.test` would just run the given test (if provided). Right?

package.json
```diff
@@ -37,7 +37,8 @@
     "test:example": "jest ./scripts/example.test.js",
     "prepublishOnly": "echo 'To publish this, go to build/ and run 'yarn publish''",
     "watch:test": "yarn test --watch",
-    "test": "NODE_ENV=test jest src/ scripts/",
+    "test": "yarn jest src/ scripts/",
+    "jest": "NODE_ENV=test jest",
```
adashst: @dan-lee Think this 👆🏻is ok?

dan-lee: Great!

PR #658, opened 2019-02-19
Fix white button color on bright backgrounds

fivenp:
This fixes the "make all buttons white" bug in elements....

We should REALLY implement a contrast color checker "soon" into elements.... I'll put it on my todo list....

image

### DIFF
--- a/src/molecules/Button.jsx
+++ b/src/molecules/Button.jsx
@@ -16,12 +16,13 @@ const baseStyle = {
 }
 
 const primary = ({
+  color,
   disabled,
   disabledColor,
   disabledBackgroundColor,
   backgroundColor,
 }) => ({
-  color: 'white',
+  color: color,
   background: disabled ? disabledBackgroundColor : backgroundColor,
 })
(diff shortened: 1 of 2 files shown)


src/molecules/Button.jsx
```diff
@@ -16,12 +16,13 @@ const baseStyle = {
...
 
 const primary = ({
+  color,
   disabled,
   disabledColor,
   disabledBackgroundColor,
   backgroundColor,
 }) => ({
-  color: 'white',
+  color: color,
```
dan-lee:
```suggestion
  color,
```

src/molecules/Button.jsx
```diff
@@ -57,7 +58,7 @@ function styles(
     ':hover': isSecondary
       ? !disabled && { color: col(backgroundColor, lightness(-20)) }
       : {
-          color: 'white',
+          color: color,
```
dan-lee:
```suggestion
          color,
```

PR #666, opened 2019-02-22
chore: resolve vulnerabilities in deps

svlapin:
chore: resolve vulnerabilities in deps and add `security-check` as travis step

Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ node_js: '8'
 cache: yarn
 before_script:
   - npm install -g yarn@1.12.3
-  - npm run security-check
+  - yarn security-check
 script:
-  - npm run test
-  - npm run lint
+  - yarn test
+  - yarn lint
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 language: node_js
 node_js: '8'
 cache: yarn
+before_script:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: node_js
 node_js: '8'
 cache: yarn
+before_script:
+  - npm install -g yarn@1.12.3
```
yamafaktory: I think that we don't need to install Yarn on Travis CI, it's available by default (https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn).

svlapin: It is, but in order to use `yarn audit` we need `>=1.12.3`, and on travis it's `1.3.2`

yamafaktory:
Alright, then that make sense 👍. However, what about using yarn anyway for
the commands instead?

svlapin:
@yamafaktory 
you mean doing `yarn security-check` instead of `npm run security-check` ? I thought about it, but decided to be consistent with `npm run test` and `npm run lint` already there.
Would you prefer to replace them with `yarn`, too?

yamafaktory:
@svlapin yes exactly! Since we are using it by default on our projects, I guess this would make sense.

PR #667, opened 2019-02-22
added extra padding on the text input field, so the checkmark doesnt …

iamn1ck:
…overlap over the text

Checklist:

### DIFF
--- a/src/atoms/Input.jsx
+++ b/src/atoms/Input.jsx
@@ -13,7 +13,7 @@ const styles = {
       boxSizing: 'border-box',
       height: 50,
       width: '100%',
-      padding: '0 15px',
+      padding: '0 35px',
       paddingTop: showLabel ? 10 : 0,
       transition: 'padding-top .225s ease-out',
       border: 0,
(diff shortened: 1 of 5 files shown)


fivenp:
uhm wait - this will „break“ all our forms!

dan-lee:
[You can see it very clearly here in storybook](https://deploy-preview-667--allthings-elements.netlify.com/?selectedKind=Forms&selectedStory=SimpleForm&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

This would break a lot of form layouts, what do you want to achieve? Please be more descriptive in your PR.

fivenp:
As daniel said - what's the expected result?

Is it maybe to bring checkbox and form input labels/placeholders to be "in line"?
image

iamn1ck:
This is before, and you can see the checkmark covers the text at the end
screenshot_2019-02-25 storybook 1

This is the after
screenshot_2019-02-25 storybook

fivenp:
well - in this case a single paddingRight: 35 will do the trick (otherwise it also affects the left one which is off...) - but:

* Only if the input really uses the checkmark

fivenp:
fixed by #811

PR #710, opened 2019-03-11
Fix: Spinner is static on IE

kialysa:
bug: on IE, `svg animate` is not supported, and `Spinner` is static (does not move)

fix: determine browser with helper function, and switch to a div with css `keyframes`

others: determine color of spinner in order of `this.props.color` →`theme.primary` → `default color`

Checklist:

JIRA: https://allthings.atlassian.net/browse/APP-2836


src/atoms/Spinner.jsx
```diff
@@ -8,52 +39,61 @@ import PropTypes from 'prop-types'
...
-    <circle
-      cx="50"
-      cy="50"
-      r="40"
-      stroke="#369980"
-      fill="none"
-      strokeWidth="8"
-      strokeLinecap="round"
+
+const Spinner = ({ color = '#369980', size = 30 }) =>
```
janrembold: please use ColorPalette from allthings/colors dependency

janrembold: hopefully there is one that fits 🤔

kialysa: good point! Asked michael, and he said ColorPalette.blue would be good 💁‍♀️
Will change.

kialysa: changed, but the auto-generated Spinner.md now says `default: ColorPalette.blue`. Think this would look odd?

janrembold: Hmmm, good question. Actually it is correct but it needs some "insides" for external developers that maybe don't know anything about ColorPalette. But I think having the right color code is worth this confusion

PR #722, opened 2019-03-14
Feature/jsx labels for typeahead

elvizcacho:


src/organisms/Typeahead.jsx
```diff
@@ -269,7 +283,13 @@ export default class Typeahead extends React.PureComponent {
         initialInputValue={defaultValue}
         initialSelectedItem={defaultSelectedItem}
         inputValue={value}
-        itemToString={item => (item ? item.label : '')}
+        itemToString={item =>
+          item
+            ? typeof item.label === 'object'
```
dan-lee: Can we use `React.isValidElement` here?

src/organisms/Typeahead.jsx
```diff
@@ -269,7 +283,13 @@ export default class Typeahead extends React.PureComponent {
...
+              ? this.getTextFromJSX(item.label)
```
dan-lee: ⚠ **I think it's problematic and smelly to do introspection on components.**
It's hard to document and tricky to ensure that components will follow this, and it could break any moment.

Maybe something like a callback function would be a better API.

src/organisms/Typeahead.jsx
```diff
@@ -288,10 +308,27 @@ export default class Typeahead extends React.PureComponent {
           selectHighlightedItem,
           toggleMenu,
         }) => {
-          const filtered = matchSorter(items, inputValue, {
+          const isObject = typeof (items[0] && items[0].label) === 'object'
```
dan-lee: Same as above: What about `React.isValidElement`?

src/organisms/Typeahead.jsx
```diff
@@ -288,10 +308,27 @@ export default class Typeahead extends React.PureComponent {
...
+
+          let itemsToBeProcessed = items
+          if (isObject) {
```
dan-lee: There's a lot of conditional `map`s and `let`s which is not so nice.
Can't we handle this in one reduce?

src/organisms/Typeahead.jsx
```diff
@@ -322,64 +359,101 @@ export default class Typeahead extends React.PureComponent {
...
+                        outline: 'none',
+                        width: '100%',
+                        height: INPUT_FIELD_HEIGHT,
+                      })}
+                    />
+                  </>
+                ) : null}
+
+                {isObject && selectedItem ? (
+                  <div
```
dan-lee: We shouldn't use `div` (use `View` instead)

src/organisms/Typeahead.jsx
```diff
@@ -322,64 +359,101 @@ export default class Typeahead extends React.PureComponent {
...
+                    className="jsx"
```
dan-lee: We shouldn't need a class name here.

src/organisms/Typeahead.jsx
```diff
@@ -322,64 +359,101 @@ export default class Typeahead extends React.PureComponent {
...
+                        color: '#000',
+                        outline: 'none',
+                        width: '100%',
+                        height: INPUT_FIELD_HEIGHT,
+                      })}
+                    />
+                  </>
+                ) : null}
+
+                {isObject && selectedItem ? (
```
dan-lee: Cant this just be 
```
{!isObject || !selectedItem
  ? ...
  : ...
}
```
?

src/organisms/__snapshots__/Typeahead.test.jsx.snap
```diff
@@ -1,5 +1,555 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Test the typeahead component Test typeahead component with jsx labels should be a simple static one - in depth testing of the core behavior with JSX labels 1`] = `
```
dan-lee: Phew. I am not so sure these snapshots help us that much. This file is now over 3000 lines long. Just imagining to look at diffs makes me anxious.

elvizcacho:
@all the build that is not passing in travis is due a service that is down at the moment that checks something regarding security

src/organisms/Typeahead.jsx
elvizcacho: I need it for the tests only

src/organisms/__snapshots__/Typeahead.test.jsx.snap
elvizcacho: It was created automatically when I run the tests, should I delete it?

src/organisms/Typeahead.jsx
dan-lee: In that case it's even worse, we shouldn't change production code just for the sake of tests. I am sure there's other ways to test for that.

src/organisms/__snapshots__/Typeahead.test.jsx.snap
dan-lee: It was added because of this:

https://github.com/allthings/elements/blob/014fe0451460d576b2ad2e53690d175797d69e48/src/organisms/Typeahead.test.jsx#L396-L400

elvizcacho: mmm... But the other tests have it as well https://github.com/allthings/elements/pull/722/files#diff-dc821506da3c26abb00bb2797d9f79f0R57

dan-lee: I know, and I think it can be an overkill to just shotgun snapshot the whole component. I think it gets harder to provide any value as the component grows.
 But we still can use it, I am not saying we should get rid of it

src/organisms/Typeahead.jsx
elvizcacho: True !A || !B = !(A && B)

src/organisms/Typeahead.jsx
elvizcacho: Yes, we can. Solved!

src/organisms/Typeahead.jsx
elvizcacho: Solved!

src/organisms/Typeahead.jsx
elvizcacho: Solved!

src/organisms/Typeahead.jsx
elvizcacho: Can you elaborate more on this topic? currently the label component has to have a property called either text or label. I wanted to be able to call a function from the React component to get the label text but I didn't find a way to do only item.label.getText()

src/organisms/Typeahead.jsx
elvizcacho: When I use view I cannot reference that container on tests

src/organisms/Typeahead.jsx
elvizcacho: Do you mean something like this?:
```javascript
<Typeahead
...
items={items}
getText={jsxLabel => jsxLabel.props.text}
...
/>
```

dan-lee: No, that would be the same introspection.

Why don't just provide a shape like this instead of doing the magic by looking into the props of an React element? Looks cleaner to me

```
{
  items: PropTypes.arrayOf(
    PropTypes.shape({
      label: PropTypes.node.isRequired,
      value: PropTypes.any.isRequired,
      icon: PropTypes.element
    })
  ).isRequired
}
```

elvizcacho: mmm ok, I'll change it to this

elvizcacho:
@dan-lee and @yamafaktory  I reimplemented it from scratch in another branch. I will close this PR

PR #724, opened 2019-03-14
Add icon to label on typeahead component

elvizcacho:

### DIFF
--- a/doc/reference/organisms/Typeahead.md
+++ b/doc/reference/organisms/Typeahead.md
@@ -24,3 +24,4 @@ data.
 |onSelect|func|Callback triggered when selecting an item.
 |placeholder|string|The placeholder displayed in the input field.
 |value|string|The value of the component, makes this a controlled component.
+|icon|element|Optional icon next to label
--- a/doc/reference/organisms/Typeahead.md
+++ b/doc/reference/organisms/Typeahead.md
@@ -14,7 +14,7 @@ and uncontrolled components.
 |placement|enum|If "top", then the list should be reversed and extended upwards, if "bottom" (default) then downwards<br>Default: 'bottom'
 |isLoading|bool|The loading state of the component, e.g when externally fetching some
 data.
-|items **(required)**|arrayOf|The items passed to component as an array of objects.
+|items **(required)**|arrayOf|The items passed to component as an array of objects. (icon is optional)
 |limit|number|The maximum number of items displayed in the menu.<br>Default: 20
 |menuHeight|number|The height of the menu in pixels.<br>Default: 300
 |onClearSelection|func|Callback triggered when clearing the selection.<br>Default: _ => _
@@ -24,4 +24,3 @@ data.
 |onSelect|func|Callback triggered when selecting an item.
 |placeholder|string|The placeholder displayed in the input field.
 |value|string|The value of the component, makes this a controlled component.
-|icon|element|Optional icon next to label
--- a/src/organisms/Typeahead.jsx
+++ b/src/organisms/Typeahead.jsx
@@ -73,6 +73,8 @@ export default class Typeahead extends React.PureComponent {
     placeholder: PropTypes.string,
     /** The value of the component, makes this a controlled component. */
     value: PropTypes.string,
+    /** Optional icon next to label */
+    icon: PropTypes.element,
   }
 
   static defaultProps = {
@@ -217,6 +219,9 @@ export default class Typeahead extends React.PureComponent {
         }),
       })}
     >
+      {item.icon ? (
+        <span {...css({ marginRight: '15px' })}>{item.icon}</span>
+      ) : null}
       <Text size="m">{item.label}</Text>
     </ListItem>
   )
@@ -292,6 +297,12 @@ export default class Typeahead extends React.PureComponent {
             keys: ['label'],
           }).slice(0, limit)
 
+          const highlightedFilteredItem = filtered[highlightedIndex]
+          const showIcon =
+            selectedItem &&
+            highlightedFilteredItem &&
+            highlightedFilteredItem.icon
+
           // Opt for <div> here because we don't want to mess with downshifts
           // getRootProps and refKey, which is kind of strange.
           return (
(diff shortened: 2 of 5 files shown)


src/organisms/Typeahead.jsx
```diff
@@ -307,112 +318,134 @@ export default class Typeahead extends React.PureComponent {
...
-                      'linear-gradient(to right, rgba(0,0,0,0) 0%,rgba(192,192,192,0) 52%,rgba(244,244,244,0) 66%,rgba(255,255,255,0.6) 81%,rgba(255,255,255,1) 88%,rgba(255,255,255,1) 100%)',
-                    bottom: 0,
-                    content: `''`,
-                    left: 0,
-                    pointerEvents: 'none',
-                    position: 'absolute',
-                    top: 0,
-                    width: '100%',
-                  },
+                  display: 'flex',
```
yamafaktory: Most of those CSS props can be directly converted as props attached to the <View /> component.

src/organisms/Typeahead.jsx
```diff
@@ -307,112 +318,134 @@ export default class Typeahead extends React.PureComponent {
...
+                  alignItems: 'center',
+                  width: '100%',
+                  backgroundColor: '#FFF',
                 })}
               >
-                <Absolute top={0} left={0} {...css({ width: '100%' })}>
+                {showIcon ? (
+                  <span
```
yamafaktory: We should just use <View /> component in this case.

src/organisms/Typeahead.jsx
```diff
@@ -307,112 +318,134 @@ export default class Typeahead extends React.PureComponent {
...
+                              inputValue,
+                              itemText:
+                                (highlightedFilteredItem &&
+                                  highlightedFilteredItem.label) ||
+                                '',
+                            })
+                          : ''
+                      }
+                      {...css({
+                        background: '#fff',
```
yamafaktory: We should use the ColorPalette instead.

src/organisms/Typeahead.jsx
```diff
@@ -307,112 +318,134 @@ export default class Typeahead extends React.PureComponent {
...
+                        border: 'none',
+                        boxShadow: 'none',
+                        color: '#999',
```
yamafaktory: Same here.

src/organisms/Typeahead.jsx
yamafaktory: Actually, you're not faulty... it was already like that before...

src/organisms/Typeahead.jsx
elvizcacho: That was already there

yamafaktory: Yeah sorry... see my previous comment :P.

src/organisms/Typeahead.jsx
elvizcacho: It was already there as well.

elvizcacho: https://github.com/allthings/elements/pull/724/files#diff-4eca93e83d29c7cb88cecf6d72e81508L345

src/organisms/Typeahead.jsx
elvizcacho: https://github.com/allthings/elements/pull/724/files#diff-4eca93e83d29c7cb88cecf6d72e81508L342

src/organisms/Typeahead.jsx
elvizcacho: https://github.com/allthings/elements/pull/724/files#diff-4eca93e83d29c7cb88cecf6d72e81508R309

elvizcacho: ?

doc/reference/organisms/Typeahead.md
```diff
@@ -24,3 +24,4 @@ data.
 |onSelect|func|Callback triggered when selecting an item.
 |placeholder|string|The placeholder displayed in the input field.
 |value|string|The value of the component, makes this a controlled component.
+|icon|element|Optional icon next to label
```
dan-lee: This belongs to the `item` shape, right?

elvizcacho: You right, my fault sorry. it was fixed already

src/organisms/Typeahead.jsx
```diff
@@ -307,112 +317,134 @@ export default class Typeahead extends React.PureComponent {
...
+                  display: 'flex',
+                  alignItems: 'center',
+                  width: '100%',
+                  backgroundColor: '#FFF',
                 })}
               >
-                <Absolute top={0} left={0} {...css({ width: '100%' })}>
+                {showIcon ? (
+                  <span
+                    id="selected-item-icon"
```
dan-lee: I'm still not a fan of "polluting" code with a random ID, which might have a CSS rule or another selector attached to it. I think with [`react-testing-library`](https://github.com/kentcdodds/react-testing-library) we would have an easier time of testing.

src/organisms/Typeahead.jsx
```diff
@@ -307,112 +317,134 @@ export default class Typeahead extends React.PureComponent {
...
+                      alignItems: 'center',
+                      marginLeft: '15px',
+                    })}
+                  >
+                    {highlightedFilteredItem.icon}
+                  </span>
+                ) : null}
+                <Relative
+                  {...css({
+                    width: showIcon ? 'calc(100% - 30px)' : '100%',
```
dan-lee: Why -30px? Can't we just use flex box?

elvizcacho: I think, it's not possible since there are floating elements. If you check what is inside that relative component, you will find there are 2 inputs that are using position: absolute.

elvizcacho: Ok, it's possible (I just tried out) but I have to duplicate the icon for the 2 inputs. I will push the new approach and you can choose what you like better.

elvizcacho: I tried to move the icon next the 2 inputs but I still need to use the calc function to adjust the width of the input when there is an icon next to it + I duplicated the icon (every input had to have it). So I think this solution here is the better.

elvizcacho: The -30px come from the margin-left: 15px from the icon component + the padding: 0 15px fro the input.

src/organisms/Typeahead.jsx
elvizcacho: But we used Ids several times on the tests. 
https://github.com/allthings/elements/pull/724/files#diff-dc821506da3c26abb00bb2797d9f79f0R13

elvizcacho: and those ids are in production code as well.

dan-lee: We should only use them if we can control it via props imo

dan-lee: Doesn't mean we should repeat the mistakes :) I only found a bad `id` in `SimpleLayout` (which is most probably a left over from copy & pasting).
`CDNIntlProvider` could've use any other selector, that's true. But it's *fairly* unique by calling it `__ELEMENTS_INTL__`. Otherwise I don't see where else elements defines the `id`.

elvizcacho: Solved!

elvizcacho: I had to use the nuclear option

dan-lee: 😁 Ok I can live with that

PR #805, opened 2019-05-09
Add dropdown input value

janrembold:
This branch 
- updates to Storybook 5.0.11
- removes import/export index.js 
- refactores all imports from index.js to standard compliant imports
- makes the Dropdown component controllable


janrembold:
@mosch please firmly look at this change: https://github.com/allthings/elements/pull/805/files#diff-eb89267f16714776a393d0f85ea63cb5R101 -> Tests were failing here and changing this resolves the tests but it feels strange to change something important.

src/organisms/Dropdown.jsx
```diff
@@ -105,16 +115,21 @@ export default class Dropdown extends React.PureComponent {
...
       items,
       initialSelectedItem,
+      selectedItem,
       placeholder,
       onSelect,
       name,
     } = this.props
     const { showScrollArrow } = this.state
+
+    console.log('render selectedItem', selectedItem)
```
mosch: Let's clean that up first 😇

src/organisms/Dropdown.jsx
```diff
@@ -154,7 +169,7 @@ export default class Dropdown extends React.PureComponent {
                       overflow: 'hidden',
                       whiteSpace: 'nowrap',
                       cursor: 'pointer',
-                      paddingRight: '30px',
+                      paddingRight: '35px',
```
mosch: Why was that changed?

stories/DropdownStory.jsx
```diff
@@ -43,6 +54,7 @@ class DropdownStory extends React.Component {
   }
 
   render() {
+    console.log('render', this.state)
```
mosch: Let's clean  that up

src/organisms/Dropdown.jsx
```diff
@@ -52,11 +57,16 @@ export default class Dropdown extends React.PureComponent {
         value: PropTypes.any.isRequired,
       })
     ).isRequired,
-    /** Initially selected */
+    /** Initially selected item - this value is uncontrolled */
```
mosch: React used the prop _defaultValue_ for this behaviour. I think it would be a good idea to make that consistent.

src/organisms/Dropdown.jsx
janrembold: Damn, too much debugging during this "little" task. Is removed now

stories/DropdownStory.jsx
janrembold: Done

src/molecules/Collapsible.jsx
```diff
@@ -1,8 +1,10 @@
 import PropTypes from 'prop-types'
 import React from 'react'
 import { css } from 'glamor'
-import { View, Icon, Text } from '../'
 import { ColorPalette, alpha } from '@allthings/colors'
+import Icon from '../atoms/Icon'
```
dan-lee: All the `import` changes shouldn't be needed anymore. At least it worked in my branch without any problems.

src/organisms/Dropdown.jsx
janrembold: The text is too long and the dots of the ellipsis effect collided with the drowdown arrow

src/organisms/Dropdown.jsx
janrembold: Don't understand. Do you mean renaming the prop `initialSelectedItem` to `defaultValue`?

src/molecules/Collapsible.jsx
janrembold: Very funny 😒Changing this took forever.
But what is the advantage of a single-import-export-file anyway?

dan-lee: It's a good indicator that the component has been exported correctly for the consumers. Happened a few times that it wasn't. Afaik it was single file imports and has only been changed lately.

src/organisms/Dropdown.jsx
janrembold: Don't know why it worked before, but now the result is still the same. I changed the total width of the input to resolve that issue:

Bildschirmfoto 2019-05-13 um 13 11 58

src/molecules/Collapsible.jsx
janrembold: For me it still doesn't work inside Dropdown element for the Icon import

Bildschirmfoto 2019-05-13 um 13 15 35

janrembold: The last error might be resolved with a direct import of Icon

dan-lee: I guess that's because `src/index.js` has been deleted in this PR.

dan-lee: Which of course is not acceptable as it would break everything 😎

janrembold: No, `src/index.js` is there. After I revert everything it still shows this error from the screenshopt above. With Icon as direct import Drowdown works fine

src/organisms/Dropdown.jsx
janrembold: This would be a major breaking change and should be done in another branch

mosch: You can keep it backwards compatible in this version and remove it in a later one.

PR #806, opened 2019-05-09
Fix storybook and audit

dan-lee:
Checklist:


src/behaviour/__snapshots__/HorizontalView.test.js.snap
```diff
@@ -83,17 +83,6 @@ exports[`HorizontalView behaviors should render non-arrays 1`] = `
         Text 2
       </Text>
     </View>
-    <View
```
dan-lee: @kialysa Suddenly the snapshots changed here. You did something in the `HorizontalView` regarding rendering children. Do you know what could've caused this sudden change?

PR #811, opened 2019-05-14
fix: Padding right on Typeahead and Input

kialysa:
adjust Typeahead and Input to not overlap icons

Checklist:


doc/reference/atoms/Input.md
```diff
@@ -21,6 +21,7 @@ TextInputs are used to allow users to enter information like names, numbers, url
 |onChange|func|Called, when the user changes something
 |value|string|The value, makes this component a controlled component
 |lines|number|Can only be used with type=text. Increase to enable multi-line input<br>Default: 1
+|paddingRight|bool|Used when there is an icon to the right of input field
```
dan-lee: We should have a more descriptive name here: Maybe `hasRightIcon`

PR #817, opened 2019-05-16
Typeahead add clear cross flag

janrembold:
Problem: 
The clear selection cross is not working as expected. When the input text gets removed the cross hides and there is no way to really clear the internal value. 

Solution/Workaround:
Added a flag to show/hide the clear cross icon in Typeahead component. The default behavior of the component is not changed by this new property. Will add a clear input value icon by myself.

### DIFF
--- a/doc/reference/organisms/Typeahead.md
+++ b/doc/reference/organisms/Typeahead.md
@@ -23,5 +23,4 @@ data.
 |onOpen|func|Callback triggered when the menu is opened.<br>Default: _ => _
 |onSelect|func|Callback triggered when selecting an item.
 |placeholder|string|The placeholder displayed in the input field.
-|showClearSelectCross|bool|Show the clear select cross inside input element.<br>Default: true
 |value|string|The value of the component, makes this a controlled component.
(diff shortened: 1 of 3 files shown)


mosch:
Would it not be better to make the clear button work by _really clearing the internal value_? 
What is that internal value and where is it stored?

janrembold:
@mosch I think refactoring the clear icon would introduce a breaking change. So this would be IMO more of a long term solution.

The clear cross is currently controlled by `selectedItem && !clearOnSelect` what seems to be not related to any clearing logic (or I just don't understand that logic behind it). 

This component has two different values:
1. The uncontrolled input value that triggers `onInputValueChange` and a possible fetch or some other logic behind it
2. The controlled `value` of the input element

* Removing the uncontrolled value is simply done by removing the text of the input.
* Removing the controlled value is not that easy, especially because the cross icon hides when you don't have any `selectedItems` 

So this PR is more of a quick fix to constantly hide the cross icon.

mosch:
@janrembold So the problem is that clicking the cross does not clear the "internal value"?

PR #821, opened 2019-05-17
set background color on input for IE11

janrembold:
Fixing the background color in Dropdown input for IE11. See screenshots below:

Before:
Bildschirmfoto 2019-05-17 um 10 07 52

After:
Bildschirmfoto 2019-05-17 um 10 08 03


janrembold:
Just found that the dropdown is broken in Firefox. The disabled input element catches click events and prevents bubbling them to the click handler in the View above (see light blue area in screenshot above). So options/items are not shown on click. 

Bildschirmfoto 2019-05-17 um 11 41 42

PR #834, opened 2019-05-27
Remove core js

mosch:
We don't need it I guess?


dan-lee:
@mosch Are you sure? Our `.babelrc` says otherwise:
https://github.com/allthings/elements/blob/66d522debb4c93426110730c772db55e7270bf3f/.babelrc#L15-L16

PR #836, opened 2019-05-28
fix: Typeahead add right padding when item is selected

kialysa:
bug: no right padding is added when an item from the dropdown is selected, leading to overlapping of text and "x" icon

fix: add right padding when item is selected

Checklist:


fivenp:
ping :)

kialysa:
@fivenp I finally did it 😂

PR #868, opened 2019-06-25
Simplify phoneinput

hasznosl:
This pull request removes every phone nr related check/restrictions from the phone input field, and 
just leaves it as a simple input field with `type: "tel"`.

Checklist:


src/molecules/PhoneInput.jsx
```diff
@@ -156,159 +18,23 @@ class PhoneInput extends Component {
...
-      this.setState({
-        numberEntered: false,
-        backspaced: false,
-        deleted: false,
-        spanZ: input.value ? 0 : 10,
-      })
-      this.props.onChange && this.props.onChange(removeFormat(formatted))
+    this.setState({
+      value: this.textInput.value,
+      spanZ: this.textInput.value ? 0 : 10,
```
dan-lee: Do we actually use `spanZ`? Couldn't find a usage

src/molecules/PhoneInput.jsx
```diff
@@ -156,159 +18,23 @@ class PhoneInput extends Component {
...
-  }
-
   handleFocus = () => {
     const { value } = this.textInput
     this.setState({ spanZ: value ? 0 : 10 })
     // time out of 0sec b/c setSelectionRange will only execute properly after focus event
     setTimeout(() => {
-      const cursorPosition = value === '+' ? 1 : value.length
-      this.textInput.setSelectionRange(cursorPosition, cursorPosition)
+      this.textInput.setSelectionRange(value.length, value.length)
```
dan-lee: For what reason do we want to keep that? Shouldn't it behave like a "normal" input field.

src/molecules/PhoneInput.jsx
dan-lee: Sorry found it, was hidden in the diff. But do we actually need it anymore?

fivenp:
one more little question... since its a phone number - we should maybe only allow numbers? (no space, +, whatsoever - only pure 0-9) 

what do you guys think?

hasznosl:
@dan-lee @fivenp please rereview this, thanks

fivenp:
One more tiny thing.... Since our `TextIntput` has an opinionated border the `PhoneInput` should also get one? :)

dan-lee:
I thought that was a mistake and we didn't want to repeat that.

mosch:
I agree, let's try to move away from the wrapping things.

PR #869, opened 2019-06-25
ButtonGroup

mosch:
Read the docs and gimme feedback.

Checklist:


src/molecules/ButtonGroup.jsx
```diff
@@ -0,0 +1,36 @@
...
+class ButtonGroup extends React.Component {
+  render() {
+    const { children } = this.props
+
+    return (
+      <View direction="horizontal">
+        {React.Children.map(children, (child, i) => (
+          <View
+            style={{
+              marginRight:
```
dan-lee: Why not something with regular CSS, like

```
css({
  '> *': {
    marginRight: 10,
  },
  '> *:last-child': {
    marginRight: 0,
  },
})

mosch: That's so much better!

PR #872, opened 2019-06-26
Version 5.0.0

mosch:
**Breaking Changes** 
- Typescript all the way - Hooray 🎉🥳👯‍♀️
- Peer dependency `React` upgraded to `^16.8` in order to use hooks.
- `Card`: Removed support for `containerStyle`, use style instead.
- `Calendar`: `onChange` now has an optional first argument on clear instead of an empty string.
- `withTheme` HOC has been dropped (use `Theme` or `useTheme` hook)
 - **Dropped refs** 
   _Note: We could/should use `forwardRef` to get them back_.
   - `View`: Changed from `onRef` to `ref` (uses `forwardRef`) 
   - `Textarea`: Removed support for `onTextarea`.
   - `RadioButton`: dropped support for `inputRef`.
- `CDNIntlProvider` was removed from `@allthings/elements` and put into its own package: [`@allthings/cdn-intl-provider`](https://github.com/allthings/cdn-intl-provider).

**Other changes**
- `ResourceProvider`: Migrated to new react context API 
- `Checkbox` renamed to `FormCheckbox`. The "new" `Checkbox` component only is a checkbox instead of being wrapped with additional styles.
- `Pill` now supports `onClickDelete`
- `Dropdown` now supports `disabled`
- `SearchableDropdown` now supports `disabled`
- `Input` now supports `disabled`
- `Text` property `strong` is now deprecated. Use `weight` instead:
`<Text weight="semi-bold" /> === <Text strong />`
- Using `const x = (props: IMyProps) => {...}` for component declarations

**New components**
- `SearchableDropdown`: A dropdown component with an additional search input for filtering items. It's a hybrid component between Dropdown and Typeahead.
 
**Todo**
- [x] Use hook over HOC/render function.
- [ ] Make Logo useful
- [x] Get rid of withTheme
- [x] Make theme work same on all components

**Upgrade notes**

`withTheme` was removed in favor for the `useTheme` hook. If you can't or don't want to rewrite everything to use hooks, you can either use the `Theme` component (Function as a child) or take a simple wrapper to restore the higher order functionality:
<details>
<summary>See higher order implementation example</summary>

(_This is API compliant, i.e. you can use it as it was used before._)

```jsx
import React from 'react'
import { useTheme } from '@allthings/elements/Theme'

export const withTheme = (
  mapThemeToProps = undefined,
) => Component => props => {
  const { theme } = useTheme()
  const themeProps = !mapThemeToProps
    ? { theme }
    : mapThemeToProps(theme, props)

  return <Component {...props} {...themeProps} />
}
```

</details>

**How to release a new beta version**

```
yarn version 5.0.0-beta.XY
yarn release
```


dan-lee:
Let's go for [`React.forwardRef`](https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components) then.

src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    const arrowDirection = placement === 'top' ? 'arrow-up' : 'arrow-down'
+
+    return (
+      <React.Fragment>
+        {showClearIcon && (
+          <Icon
+            color="black"
+            data-testid="searchable-dropdown-clear-icon"
```
mosch: That's the first time we do this. I'm not convinced yet that this is the right thing to do. But if so – we should be consistent about this.

mosch:
@janrembold can you add some changelog for the `SearchableDropdown` in the first comment?

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: I used the react-testing-library for testing and firing events to this component. This library encourages the use of `data-testid` for simple selection of elements. In cockpit we use `data-e2e` but it doesn't seem to be possible to select those attributes with react-testing-library. 

Or should we better use something like Enzyme for testing? What is your preference @mosch / @dan-lee?

dan-lee: I am definitely pro react-testing-library as we can test the actual output (i.e. HTML). Also we should not hardcode either `data-e2e` or `data-testid`, this should be done by the implementer side. Maybe we need to spread rest props for that (like in many other places in elements)

Enzyme uses internals of React and always needs to keep up with the development of React (we had a lot of issues in the past because it was out of sync)

janrembold: I would agree if that would work for complex components like, in this case, the SearchableDropdown. Here we need to select different places inside the component. Only spreading rest props e.g. on the outer container is not sufficient.

mosch:
`add custom scrollbar for webkit` @janrembold @fivenp not happy about this!

src/Input/Input.tsx
```diff
@@ -0,0 +1,355 @@
...
+      boxSizing: 'border-box',
+      height: 50,
+      width: '100%',
+      padding: '0 15px',
+      paddingLeft: translated && 40,
+      paddingTop: showLabel ? 10 : 0,
+      paddingRight: paddingRight ? 50 : 15,
+      transition: 'padding-top .225s ease-out',
+      border: 0,
+      ':required': {
```
mosch: @DmitryMarkov your changes are here in this branch now

PR #887, opened 2019-07-12
Add support for noItemsText in Dropdown

mosch:
New Feature:
- Dropdown: New prop "noItemsText" allows you to show a unselectable text if no items are available to choose.

Screenshot 2019-07-12 at 10 07 32

Checklist:


fivenp:
hmmmm...... this is a little paradox :)

why would a user be able click on a dropdown at all if it has no items?

if this is intended to be the groundwork for a filterable dropdown it makes of course sense in that context :)

mosch:
> why would a user be able to click on a dropdown at all if it has no items?

We have the situation where do should choose something for complete the form – but there is nothing you can choose from for some reason. What would you suggest as a solution?

janrembold:
@mosch I don't know the exact use-case but showing a simple text or even better show nothing at all sounds much better than adding a dropdown that tells the user "it was useless that you click me". What I mean is that a Dropdown forces the user to make a click just to inform her that there is nothing to choose. So I would propose to completely hide it -> KISS

mosch:
Ok ☹️

PR #888, opened 2019-07-15
App 3515 searchable dropdown

janrembold:
⚠️ This PR depends on the `typescripted` branch

Checklist:


src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,327 @@
...
+  noResultsText,
+  onLoadMore = noop,
+  onSearch,
+  onSelect,
+  placeholder = '',
+  placement = EDropdownDirection.BOTTOM,
+  searchTerm = '',
+  selectedItem,
+}: ISearchableDropdownProps) => {
+  const searchRef: RefObject<HTMLInputElement> = createRef()
```
dan-lee: Should rather be: 
```suggestion
  const searchRef = useRef<HTMLInputElement>()
```

Right?

janrembold: 👍  Done

src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,326 @@
...
+  /** Callback triggered when clicking on "Load more" in items list */
+  onLoadMore?: () => void
+  /** Callback triggered when clearing the selection. */
+  onSelect: (item: IDropdownItem) => void
+  /** Callback triggered when search value changes */
+  onSearch?: (event: SyntheticEvent) => void
+  /** The placeholder displayed in the input field. */
+  placeholder?: string
+  /** If "top", then the list should be reversed and extended upwards, if "bottom" (default) then downwards */
+  placement?: EDropdownDirection
```
dan-lee: Can we use `placement?: 'top' | 'bottom'` here? The docs only show `EDropdownDirection` otherwise.

src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,326 @@
...
+  name = '',
+  noResultsText,
+  onLoadMore = noop,
+  onSearch,
+  onSelect,
+  placeholder = '',
+  placement = EDropdownDirection.BOTTOM,
+  searchTerm = '',
+  selectedItem,
+}: ISearchableDropdownProps) => {
```
dan-lee: To be more consistent we should use: `const SearchableDropdown: FunctionComponent<ISearchableDropdownProps>`.

src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,326 @@
...
+
+  const handleClearIconClick = (
+    event: SyntheticEvent,
+    clearSelection: () => void,
+  ) => {
+    event.stopPropagation()
+    clearSelection()
+  }
+
+  const stateReducer = (
```
dan-lee: I expect this component to have a lot of re-renders, would it make sense to use `useCallback`? Or would it be premature optimization, not sure... (also for the following callback function)

stories/SearchableDropdownStory.tsx
```diff
@@ -0,0 +1,340 @@
...
+  const [items, setItems] = useState()
+  const [searchQuery, setSearchQuery] = useState('')
+  const [loadMore, setLoadMore] = useState('')
+  const [itemCount, setItemCount] = useState(ITEMS_LIMIT)
+
+  useEffect(() => {
+    queryItems()
+  }, [searchQuery, itemCount])
+
+  const queryItems = () => {
```
dan-lee: Why not put this into `useEffect`? Doesn't really make sense to separate it.

src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -0,0 +1,326 @@
...
+          name={arrowDirection}
+          size={10}
+        />
+      </React.Fragment>
+    )
+  }
+
+  const renderSearchInput = () => (
+    <Input
+      ref={searchRef as any}
```
dan-lee: Why `as any`? Can you try `const searchRef = useRef<HTMLInputElement>(null)`, maybe it's a similar problem I had before.

stories/SearchableDropdownStory.tsx
janrembold: The reason for a separation in this case is the readability and maintainability of the useEffect method. If something is added here in the future it will be hard to divide between different functionalities. I don't think an additional method doesn't hurt here anyways

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: 👍 done

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: I think using enums/interfaces where ever possible is a good thing even for such small structures like `'top' | 'bottom'`. It's also good to be consistent with other projects (e.g. Cockpit) where we heavily use enums and interfaces.

But IMHO the interesting questions are:
- how is this enum communicated in the docs?
- are those enums exported and usable in external projects (are they available in the build package)?

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: Good point, but I think this would be a premature optimization. The component is pretty lightweight, because all the heavy lifting is done outside (searching, pagination / load more logic...). I also thought about integrating a debouncer but left that out because the outer search callback should handle that if search re-renders result in a performance problem

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: Nice catch! I've missed the `null` parameter in the `useRef` method. Using any was my lazyness..shame on me :(

src/SearchableDropdown/SearchableDropdown.tsx
dan-lee: `type Placement = 'top' | 'bottom'` would show correctly in the docs.

stories/SearchableDropdownStory.tsx
dan-lee: You also need the variables/functions in the dependency array you are actually using in `useEffect` so you also need to add `queryItems` otherwise it wouldn't update. See [`eslint-plugin-react-hooks
`](https://github.com/facebook/react/tree/master/packages/eslint-plugin-react-hooks)

_(I guess this also requires some reordering)_

dan-lee: Also good insights about these can be found here: https://overreacted.io/a-complete-guide-to-useeffect/#dont-lie-to-react-about-dependencies

dan-lee: I say all of this because now we have transitive dependencies listed. Every variable which changes in `queryItems` should be in the dependency array, the lint plugin can help with this, but if we move it outside it doesn't help as much.

src/SearchableDropdown/SearchableDropdown.tsx
janrembold: Ok, why not...

stories/SearchableDropdownStory.tsx
janrembold: Thanks for the insights! Moved it into useEffect 👍

PR #909, opened 2019-07-30
Searchable dropdown v2

janrembold:
- Remove babel noop import
- add background color to dropdown wrapper
- minor change in story
- update snaps


src/SearchableDropdown/SearchableDropdown.tsx
```diff
@@ -38,19 +34,16 @@ const styles = {
...
-        : ColorPalette.white,
-      color: 'black',
-      overflow: 'hidden',
-      paddingRight: '30px',
-      pointerEvents: 'none',
-      textOverflow: 'ellipsis',
-      whiteSpace: 'nowrap',
-      width: 'calc(100% - 5px)',
-    }),
+  disabledInput: css({
```
mosch: Do we still have a visual indication that the dropdown is disabled?

mosch:
We should not reformat the markdown files since they are generated that way and the next one would undo that change

PR #911, opened 2019-07-31
Resolves APP-3810: IE/App: typeahead needs some fixin’

egeogretmen:
There is also another rule in cockpit css about IE quirks. 
https://github.com/allthings/cockpit/blob/master/public/static/css-src/style.css#L46
https://developer.mozilla.org/en-US/docs/Web/CSS/%3A%3A-ms-reveal

It is not in the scope of this PR, but maybe that should also be added?
Checklist:

### DIFF
--- a/src/Input/Input.tsx
+++ b/src/Input/Input.tsx
@@ -42,9 +42,6 @@ const styles = {
       '&:-webkit-autofill': {
         paddingTop: '10px !important',
       },
-      '::-ms-clear': {
-        display: 'none',
-      },
     }),
   area: (lines: number, showLabel: boolean) =>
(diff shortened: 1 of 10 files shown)


src/Input/Input.tsx
```diff
@@ -39,6 +39,9 @@ const styles = {
       '&:-webkit-autofill': {
         paddingTop: '10px !important',
       },
+      ':-ms-clear': {
```
dan-lee: Shouldn't this be `::-ms-clear`?

egeogretmen: yes that's right, thanks!

PR #926, opened 2019-09-02
Allow to disabled Input and Dropdown

mosch:
Setting the Dropdown or Input to _'disabled'_ did not change anything visually. It shows a different cursor now.

Checklist:


dan-lee:
Oh boy, that was definitely missing! What about also adding a grey background, which is basically de-facto standard anyway?

fivenp:
We could theoretically add a `ColorPalette.whiteIntense` as background color if it's disabled.... But we would need to do so for all the disabled input/form elements! (`Input`, `Dropdown`, 'SearchableDropdown`, etc)

mosch:
@dan-lee Can we just merge it for now? I'd like to use it – if we like to improve it further, let's open a new PR.

PR #928, opened 2019-09-09
Add spacing helper

dan-lee:
Does this make sense to add?

Example usage:

```jsx
function App() {
  return (
    <div
      {...spacing.inner({ vertical: 30 })}
      {...spacing.outer(15)}
    >
      Content
    </div>
  )
}
```

Checklist:

### DIFF
--- a/src/Spacer/Spacer.tsx
+++ b/src/Spacer/Spacer.tsx
@@ -1,6 +1,31 @@
 import React from 'react'
 import View from '../View'
-import { css } from 'glamor'
+import { css, StyleAttribute } from 'glamor'
+import isNumeric from '../utils/isNumeric'
+
+type SpacingType = number | { horizontal?: number; vertical?: number }
+
+interface ISpacing {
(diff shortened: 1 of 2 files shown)


mosch:
Great idea! Maybe it's worth thinking it spacing steps like `1px 3px 5px 10px 15px`?

fivenp:
> Great idea! Maybe it's worth thinking it spacing steps like `1px 3px 5px 10px 15px`?

being more verbose makes sense somehow.... we usually use 5px steps for margins/paddings (at least in 90% of the use cases) - maybe something like this? 

```
{
  "s" => 5,
  "m" => 10,
  "l" => 15,
  "xl" => 20,
}
```

dan-lee:
I think it would be sensible to provide these as defaults.

Like:
```
export const SPACING = {
  s: 5,
  m: 10,
  l: 15,
  xl: 20,
}
```

fivenp:
👑 !

PR #930, opened 2019-09-12
Switch React import to * as React

xeoxx:
This change will allow the use of `"allowSyntheticDefaultImports": true` in tsconfig.json for dependent projects.
Without this change Interface evaluation uses empty (`{}`) React.HTMLElement interfaces for property evaluation. This causes e.g. `onClick` to not be recognized as a valid property on ListItem.

Checklist:


xeoxx:
> What about https://github.com/allthings/elements/blob/typescripted/tsconfig.json#L4? Can't we switch it to `false` too?

Unfortunately I was not able to resolve:
`import Mitt, { Handler } from 'mitt'`
`import json2mq from 'json2mq'`
`import matchSorter from 'match-sorter'`
None of them have a default export, so `import * as Mitt from 'mitt'` does not work.
If anyone has a workaround for that we could change `allowSyntheticDefaultImports=false`.

yamafaktory:
@xeoxx What about just going for the ol'school:
```ts
 return `@media ${require('json2mq')(devices.map(device => MQ[device]))}`
```
https://github.com/allthings/elements/blob/5d9cf471209a90cd5cf7b199bc1c2629f85b909c/src/behaviour/Responsive.jsx#L17

mosch:
I don't see why we shouldn't do this, however:
> This causes e.g. onClick to be not recognized as a valid property on ListItem.
Where do you see this happening?

xeoxx:
> @xeoxx What about just going for the ol'school:
> ```ts
>  return `@media ${require('json2mq')(devices.map(device => MQ[device]))}`
> ```

mosch:
Just realized that there are a bunch of comma changes – please make sure that they respect our linting rules.

doc/using-themes.md
```diff
@@ -33,12 +34,13 @@ import { withTheme } from '@allthings/elements/behaviours/ThemeProvider'
...
 ```
-Simply wrap a component with *withTheme* and a property theme gets passed:
+
+Simply wrap a component with _withTheme_ and a property theme gets passed:
 
 ```js
 import { withTheme } from '@allthings/elements/behaviours/ThemeProvider'
 
-const Component = ({ theme }) => (
+const React.Component = ({ theme }) => (
```
dan-lee: Probably shouldn't be replaced

xeoxx: Reverted all `.md` changes

xeoxx:
> @xeoxx What about just going for the ol'school:
> 
> ```ts
>  return `@media ${require('json2mq')(devices.map(device => MQ[device]))}`
> ```

I'd prefer not to do that, require makes it loose the type.

PR #931, opened 2019-09-13
added support for 200 and 800 font weight

iamn1ck:
Checklist:


fivenp:
@mosch - would you mind guiding us here? :)

mosch:
`https://github.com/allthings/elements/blob/master/src/atoms/Text.jsx#L75`

we need to add a prop like `weight: PropTypes.oneOf(['normal', 'strong', 'stronger', 'light', 'lighter'])`. We could keep the strong prop to not break the API and map it to the new weight prop.

dan-lee:
Has been solved in https://github.com/allthings/elements/pull/942

PR #932, opened 2019-09-13
fix security dependency

iamn1ck:
Checklist:


package.json
```diff
@@ -27,7 +27,8 @@
     "terser": "3.14",
     "lodash": ">=4.17.2",
```
DmitryMarkov: Maybe it's better to update lodash also to 4.17.15?

PR #933, opened 2019-09-20
fix: [Listspinner] pass color down to Spinner

kialysa:
`<ListSpinner color={COLOR}/>` turns the `Circle` (aka background) into the color passed down.
Expected behavior is for the `Spinner` to be colored.

Checklist:

### DIFF
--- a/src/atoms/Spinner.jsx
+++ b/src/atoms/Spinner.jsx
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 import { css, keyframes } from 'glamor'
 import { withTheme } from '../behaviour/ThemeProvider'
 import { isIE11 } from '../utils/viewport'
+import Theme from '../behaviour/Theme'
 
 const spin = keyframes('load', {
   '0%': {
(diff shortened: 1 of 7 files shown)


src/molecules/List/ListSpinner.jsx
```diff
@@ -1,19 +1,20 @@
...
+import { number, string } from 'prop-types'
 import View from '../../atoms/View'
 import Spinner from '../../atoms/Spinner'
 import Circle from '../../atoms/Circle'
 
 /** It's a spinner for a list **/
-const ListSpinner = ({ radius = 40, ...props }) => (
+const ListSpinner = ({ color, radius = 40, ...props }) => (
   <View direction="row" alignH="center">
     <Circle radius={radius} color="white" {...props}>
```
janrembold: This color is still fixed while the other is variable. Won't this lead to an issue depending on the color combination?

src/molecules/List/ListSpinner.jsx
```diff
@@ -1,19 +1,20 @@
...
-      <Spinner size={radius - 10} />
+      <Spinner color={color} size={radius - 10} />
     </Circle>
   </View>
 )
 
 ListSpinner.propTypes = {
+  color: string,
```
janrembold: Setting the color as a required prop introduces a breaking change, right? Wouldn't it be better to make it optional and use the default color of the spinner as default value?

kialysa: this PR is for master, and is not typescripted yet. if it's required, wouldn't it be `color: string.isRequired`?

src/molecules/List/ListSpinner.jsx
kialysa: The `Circle` for listSpinner should be fixed to white, as far as i understand 🤔  The color that is passed down through props should be directed at `Spinner`, rather than `Circle`.

src/molecules/List/ListSpinner.jsx
kialysa: and I will also create the same PR for `typescripted`, as Michael mentioend  I should do above 🙂

src/atoms/Spinner.jsx
```diff
@@ -43,49 +44,57 @@ const styles = (color, size) => ({
...
+          xmlns="http://www.w3.org/2000/svg"
+          viewBox="0 0 100 100"
+          preserveAspectRatio="xMidYMid"
+        >
+          <rect x="0" y="0" width="100" height="100" fill="none" />
+          <circle
+            cx="50"
+            cy="50"
+            r="40"
+            stroke="#ffffff"
```
fivenp:
```suggestion
            stroke={colorize(background)}
```

? :) 
not 100% sure tho

src/molecules/List/ListSpinner.jsx
```diff
@@ -1,19 +1,25 @@
...
+        <Circle radius={radius} color="white" {...props}>
+          <Spinner color={colorize(color)} size={radius - 10} />
+        </Circle>
+      </View>
+    )}
+  </Theme>
 )
 
 ListSpinner.propTypes = {
+  color: string,
```
fivenp: Please make sure to also add the background color here

src/molecules/List/ListSpinner.jsx
fivenp: the LIST spinner could potentially be the `contrast` color (which is right now `white` and makes sense)

janrembold: OK great

src/molecules/List/ListSpinner.jsx
janrembold: Sorry, I didn't check that string was imported from `prop-types`. Haven't worked on non-TS code that much lately

src/molecules/List/ListSpinner.jsx
kialysa: ok! by backgroudColor means for the `Circle`, right?

src/molecules/List/ListSpinner.jsx
kialysa: ok, thanks for the feedback you two!

src/molecules/List/ListSpinner.jsx
fivenp: both - the spinner and circle

kialysa: the spinner can have a backgroundColor? Oh, you mean the "path/track" that the color is circling?

src/molecules/List/ListSpinner.jsx
kialysa: yeah me too 😅I nearly forgot we used to do that..

src/molecules/List/ListSpinner.jsx
fivenp: exactly :)

PR #934, opened 2019-09-20
chore: removing 'import * as React' and const x: React.FC<Prop>

xeoxx:
In accordance to # 11 and # 20 of the following:
https://medium.com/@martin_hotell/10-typescript-pro-tips-patterns-with-or-without-react-5799488d6680#78b9

This changes:
1. `import * as React from 'react'` to `import React from 'react'`
2. `const x: React.FC<IXProps> = (props)=>...` to `const x = (props: IXProps)=>...`

Checklist:


mosch:
👏🏼

dan-lee:
So are all `children` now type `any`?

mosch:
> So are all `children` now type `any`?

I thought the same during my review, however I'd expected the linting to complain about _implicit any_.

xeoxx:
> children

They most of the time inherit from `IViewProps` or are declared by `React.PropsWithChildren<IMyProps>`.
Unless defined otherwise they are of type `React.ReactNode`.
This is the same as when they would use `const x: React.FC<IMyProps> = ...`

dan-lee:
Generally I like the PR, but as you said now we're basically using `PropsWithChildren`, which doesn't really make a difference.
The article states that one should explicitly type their children.

xeoxx:
> Generally I like the PR, but as you said now we're basically using `PropsWithChildren`, which doesn't really make a difference.
> The article states that one should explicitly type their children.
Yap, but maybe not in this PR?

PR #941, opened 2019-10-09
Allow numeric font sizes

dan-lee:
Checklist:

### DIFF
--- a/src/Text/Text.tsx
+++ b/src/Text/Text.tsx
@@ -13,7 +13,15 @@ if (typeof window !== `undefined`) {
   ])
 }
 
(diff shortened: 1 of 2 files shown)


src/Text/Text.tsx
```diff
@@ -40,7 +48,7 @@ export const createTextStyles = ({
   block = false,
   italic = false,
   strong = false,
-  size = 'm',
+  size = 'l',
```
dan-lee: This is was changed so it matches the default props of `Text`.

dan-lee: https://github.com/allthings/elements/blob/d272e2e03664e9ddbd6c2bbb6702baa19e67e4b3/src/Text/Text.tsx#L96-L108

PR #942, opened 2019-10-09
Introduce `weight` on `Text` element and deprecate strong

dan-lee:
Checklist:

### DIFF
--- a/src/Text/Text.tsx
+++ b/src/Text/Text.tsx
@@ -65,7 +65,7 @@ export const createTextStyles = ({
     display: block ? 'block' : 'inline',
     fontFamily: '"Open Sans", Helvetica, Arial, sans-serif',
     fontStyle: italic && 'italic',
-    fontWeight: weight ? textWeightMap[weight] : strong && '600',
+    fontWeight: weight ? textWeightMap[weight] : strong ? 600 : 400,
     fontSize: typeof size === 'number' ? size : availableSizes[size],
     textDecoration:
       (underline && 'underline') || (lineThrough && 'line-through'),
--- a/src/Text/Text.tsx
+++ b/src/Text/Text.tsx
@@ -7,24 +7,12 @@ import View, { IViewProps } from '../View'
 if (typeof window !== `undefined`) {
   neue.load([
     {
-      families: ['Open Sans:n4,n4i,n6,n6i,n7,n7i'],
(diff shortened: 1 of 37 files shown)


src/Text/Text.tsx
```diff
@@ -58,7 +72,7 @@ export const createTextStyles = ({
     display: block ? 'block' : 'inline',
     fontFamily: '"Open Sans", Helvetica, Arial, sans-serif',
     fontStyle: italic && 'italic',
-    fontWeight: strong && '600',
+    fontWeight: weight ? textWeightMap[weight] : strong && '600',
```
fivenp: maybe add a last fallback if no `weight` and no `strong` is passed at all -> `400`

dan-lee: Thanks, good suggestion!

stories/SimpleLayoutStory.tsx
```diff
@@ -50,9 +50,7 @@ class SimpleLayoutStory extends Component<
               </View>
             )}
             {text && (
-              <Text strong block>
-                Pull down to load a new Text.{' '}
-              </Text>
+              <Text weight="semi-bold">Pull down to load a new Text. </Text>
```
fivenp: no `block` anymore?

src/Text/Text.tsx
```diff
@@ -7,8 +7,9 @@ import View, { IViewProps } from '../View'
 if (typeof window !== `undefined`) {
   neue.load([
     {
-      families: ['Open Sans:n4,n4i,n6,n6i'],
-      css: '//fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i',
+      families: ['Open Sans:n4,n4i,n6,n6i,n7,n7i'],
+      css:
+        '//fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i,700,700i',
```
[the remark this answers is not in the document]
fivenp: ['Open Sans:n3,n3i,n4,n4i,n6,n6i,n7,n7i']

stories/SimpleLayoutStory.tsx
dan-lee: `block = true` is the default :)

mosch:
Perfect – we should add this to the changelogs!

PR #954, opened 2019-11-26
Add `required` for Dropdown

dan-lee:
This adds a `required` prop for the `Dropdown` component.
It uses a trick to achieve this with standard HTML5 validation:
First `readOnly` and `disabled` needed to be removed as these skip validation entirely.
Then this code achieves that the caret is not visible and no keystrokes can be made:

https://github.com/allthings/elements/blob/37709b664b2ae96cc585c0a90a269faae423a0e0/src/Dropdown/Dropdown.tsx#L174-L179

Also I fixed the behaviour using a `Dropdown` inside a form. When trying to get the value it actually returned the label (i.e. the representation), so I added an hidden input which contains the value:

https://github.com/allthings/elements/blob/37709b664b2ae96cc585c0a90a269faae423a0e0/src/Dropdown/Dropdown.tsx#L192-L197

Checklist:


src/Form/Form.tsx
```diff
@@ -124,7 +124,9 @@ const Form = ({
     const elements = target.elements as HTMLFormControlsCollection
     for (let i = 0; i < elements.length; i++) {
       const element = elements.item(i) as HTMLInputElement
-      data[element.name] = element.value
+      if (element.name) {
+        data[element.name] = element.value
+      }
```
dan-lee: input fields with no `name` prop would be in this data object:

```
{
  "field": "value",
  "": "some other value"
}

```

src/Dropdown/Dropdown.tsx
```diff
@@ -59,8 +59,8 @@ export interface IDropdownProps {
   /** Disable the dropdown */
   readonly disabled?: boolean
   /** If "top", then the list should be reversed and extended upwards, if "bottom" (default) then downwards */
-  readonly placement?: 'top' | 'bottom'
-  readonly items: ReadonlyArray<IDropdownItem>
+  readonly placement?: keyof typeof Placement
```
hasznosl: 🎉

PR #977, opened 2019-12-06
fix: label for input with set defaultValue

xeoxx:
Checklist:
Fixes that the label is not displayed if an non-empty defaultValue is used on an input.

### DIFF
--- a/src/Input/Input.tsx
+++ b/src/Input/Input.tsx
@@ -242,18 +242,17 @@ const Input = ({
 }: IInputProps) => {
   const internalRef = useRef<HTMLTextAreaElement & HTMLInputElement>(null)
   const isTextArea = lines !== 1
-  const [changedValue, setChangedValue] = useState<
-    string | string[] | undefined
-  >(undefined)
+  const [value, setValue] = useState(
+    typeof props.value !== 'undefined' ? props.value : props.defaultValue,
+  )
 
-  const currentValue =
-    typeof changedValue !== 'undefined'
-      ? changedValue
-      : typeof props.value !== 'undefined'
-      ? props.value
-      : props.defaultValue
+  useEffect(() => {
+    if (typeof props.value !== 'undefined') {
+      setValue(props.value)
+    }
+  }, [props.value])
 
-  const length = typeof currentValue === 'string' ? currentValue.length : 0
+  const length = typeof value === 'string' ? value.length : 0
   const labelVisible = length > 0
   const showLabel = !!(label && labelVisible)
 
@@ -280,7 +279,7 @@ const Input = ({
   const handleChange = useCallback(
     (e: ChangeEvent<HTMLInputElement> | ChangeEvent<HTMLTextAreaElement>) => {
       setValidity(e.target, customValidity)
-      setChangedValue(e.target.value)
+      setValue(e.target.value)
       props.onChange && props.onChange(e)
     },
     [customValidity, props],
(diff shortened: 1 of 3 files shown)


src/Input/Input.tsx
```diff
@@ -240,15 +240,22 @@ const Input = ({
...
   const isTextArea = lines !== 1
-  const [value, setValue] = useState('')
-  const [length, setLength] = useState(
-    (typeof props.value === 'string' && props.value.length) || 0,
-  )
-  const currentValue = (props.value || value) as string
-  const labelVisible = currentValue.length > 0
-  const showLabel = !!(label && currentValue.length > 0)
+  const [changedValue, setChangedValue] = useState<
+    string | string[] | undefined
```
svlapin: What's the case for `string[]`?

src/Input/Input.tsx
```diff
@@ -240,15 +240,22 @@ const Input = ({
...
+  >(undefined)
+
+  const currentValue =
+    typeof changedValue !== 'undefined'
+      ? changedValue
+      : typeof props.value !== 'undefined'
+      ? props.value
+      : props.defaultValue
```
svlapin: `defaultValue` should take effect only on initial mount

src/Input/Input.tsx
```diff
@@ -240,15 +240,22 @@ const Input = ({
...
-  const [value, setValue] = useState('')
-  const [length, setLength] = useState(
-    (typeof props.value === 'string' && props.value.length) || 0,
-  )
-  const currentValue = (props.value || value) as string
-  const labelVisible = currentValue.length > 0
-  const showLabel = !!(label && currentValue.length > 0)
+  const [changedValue, setChangedValue] = useState<
+    string | string[] | undefined
+  >(undefined)
```
svlapin: I'd prefer to init with `props.value || props.defaultValue`; for the case of controlled component - `useEffect` with a dependency on `props.value` which updates `changedValue` accordingly.
In that case, the whole block `const currentValue = ...` might be redundant.

src/Input/Input.tsx
xeoxx: These are the data types an input can produce... The string[] is most likely from checkbox.

src/Input/Input.tsx
xeoxx: Yes. If there was any change, changedValue is no longer typeof undefined.

src/Input/Input.tsx
xeoxx: This would actually have an impact on the behavior of the input. Right now the init value does not trigger a validation. Only input changes trigger validation.

src/Input/Input.tsx
svlapin: How a type of value checkbox holds is related?

`changedValue` is only used within that component, and `setChangedValue` is only called with `e.target.value` in `handleChange`, where `target` is either `HTMLInputElement` or `HTMLTextAreaElement` - both results in `string`.

src/Input/Input.tsx
svlapin: Don't see how would it impact validation - `setValidity` is called exactly in `handleChange`; initializing state var `changedValue` (or updating it from changed `props.value`) won't change the fact that
> Only input changes trigger validation

src/Input/Input.tsx
svlapin: That's an implicit logic that makes the whole block pretty hard to read (related to https://github.com/allthings/elements/pull/977/files#r357135991)

PR #1040, opened 2020-02-13
fix(app-5278): overflowing input labels don't cover value anymore

xeoxx:
Fixes a bug where inputs with a long label where overflowing into the input field itself covering the input text.

Checklist:

### DIFF
--- a/src/Input/Input.tsx
+++ b/src/Input/Input.tsx
@@ -1,3 +1,4 @@
+import { ColorPalette } from '@allthings/colors'
 import { css } from 'glamor'
 import React, {
   AllHTMLAttributes,
@@ -93,10 +94,23 @@ const styles = {
   label: (translated: boolean) =>
     css({
       position: 'absolute',
-      left: translated ? 40 : 15,
+      padding: `0 15px 0 ${translated ? 40 : 15}px`,
+      width: '100%',
       fontSize: 10,
       opacity: 0,
       transition: 'all .225s ease-out',
+    }),
+  labelText: (labelVisible: boolean) =>
+    css({
+      minWidth: 0,
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      ':hover': {
+        textOverflow: 'clip',
+        whiteSpace: 'normal',
+        backgroundColor: labelVisible ? ColorPalette.white : null,
+      },
     }),
   placeholder: css({
     position: 'absolute',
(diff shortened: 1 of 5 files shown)


src/Input/Input.tsx
```diff
@@ -93,11 +94,24 @@ const styles = {
...
-      left: translated ? 40 : 15,
+      padding: `0 15px 0 ${translated ? 40 : 15}px`,
+      width: '100%',
       fontSize: 10,
       opacity: 0,
       transition: 'all .225s ease-out',
     }),
+  labelText: (labelVisible: boolean) =>
+    css({
+      minWidth: 0,
```
svlapin: Does that make sense?

src/Input/Input.tsx
```diff
@@ -93,11 +94,24 @@ const styles = {
...
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      ':hover': {
+        textOverflow: 'clip',
```
svlapin: What's the point here? Since parent is `overflow: hidden` anyway that would only show 1 or 2 additional symbols which is hardly helpful if label is too long.

src/Input/Input.tsx
```diff
@@ -93,11 +94,24 @@ const styles = {
...
+        whiteSpace: 'normal',
+        backgroundColor: labelVisible ? ColorPalette.white : null,
```
svlapin: Does that work fine if surrounding background is not white?

src/Input/Input.tsx
xeoxx: Yes, this is required in order to allow the div to collapse.

src/Input/Input.tsx
xeoxx: `hover` also changes `whiteSpace: 'normal'`. Overall it will present the label on top of the value, with a white background. As we are doing `textOverflow: 'clip'` it won't overflow in the x direction, and in the y direction it should still be hidden.

src/Input/Input.tsx
xeoxx: On `hover` it will just show the label in white, similar to `title`, but instead of a yellow title you get the text in the place where where the label would be if it was not overflowing

src/Input/Input.tsx
svlapin: Quite sure it does not - labelText is a block-level element (Text) that is rendered inside another block-level element (View), so the former takes spans the whole width of the parent (not exceeding max width), regardless its content.

xeoxx: As discussed, `view` is flex not block

PR #1118, opened 2020-07-14
fix autocomple in typeahead

elvizcacho:
Checklist:


yamafaktory:
For the record, this PR aims at fixing the autocompletion set to off not being enough in some browsers, see https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion:

> If a browser keeps on making suggestions even after setting autocomplete to off, then you have to change the name attribute of the input element.

PR #1210, opened 2021-09-23
feat: migrate to react 17

tverdohleb:
Checklist:


DmitryMarkov (commented):
I assume we need to use updated `react-motion` library and bump major version

